PR #17, opened 2015-10-23
Third party themes

hadim:


xaiki:
don't remove lang, we'll get that as soon as @insertc0in comes back up.

hadim:
ok

hadim:
Should be ok now.

PR #23, opened 2015-10-24
Update npm dependency versions.

matthewbauer:
Some of the dependencies to git weren't working for me.


xaiki:
delaying this one for a bit, i'm pretty sure it'll break things.

PR #25, opened 2015-10-24
Fix settings template.

matthewbauer:
Settings template had errors when the Trakt or TVShowApi was
disconnected.


hadim:
It works for me. Can be merged !

hadim:
I don't see tv show time settings. Is that excepted ?

xaiki:
ok, actually we're going to need a bit more than that, do you think you can work on having this go modlar, i.e. move this to the provider's code ?

PR #27, opened 2015-10-24
Update Gruntfile.js

teddy-error:
This update is necessary in order to successfully run on Windows

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -228,7 +228,7 @@ module.exports = function (grunt) {
 
         exec: {
             win: {
-                cmd: '"build/cache/win/<%= nwjs.options.version %>/nw.exe" .'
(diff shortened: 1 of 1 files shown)


xaiki:
the build part looks good, the hardcode of win64 will probably break win32

teddy-error:
To be fair, both win32 and win64 are broken currently.

MrAdder:
I was able to build win64 and win32 without this fix

wvanbreukelen:
That funny, I experience the same issues as @teddy-error on a win64 system.

MrAdder:
Yeah I know what I did I didnt use grunt start :P I went direct to the build dir and tested from there

paulhobbel:
I'm also able to build the project without that fix

xaiki:
uneeded.

PR #46, opened 2015-10-24
Missing "es" translations.

rob3ns:


xaiki:
now i remember why we splited out transitions to it's own repository... @insertc0in any ideas ?

insertc0in:
This particular language, like many others, is still being translated on Transifex.
One of the reasons why it is missing those sentences seen in this commit is because no one translated them yet on Transifex, and because the translations originally ported to Butter from Popcorn Time weren't exactly the most recent. Nothing to do with not having it's own repo.
@xaiki not sure, I got no problem as it is.

PR #50, opened 2015-10-24
Move nodecast-js dependency to npm.

matthewbauer:
nodecast-js needs to be installed via npm or else the build files won't
be available. Looking at the repository there haven't been any changes
since 0.1.2 which is currently available from npm.


MrAdder:
Works in dev enviroment Merged

PR #66, opened 2015-10-25
Butter/master

metindinccelik:
I want to contribute butter project, I have already fixed a small issue about maximize/minimize


MrAdder:
This commit would back track what we have already done

PR #75, opened 2015-10-26
Updated it translation synced with latest en.json

DarioZ98:


MrAdder:
This would send us backwards merged closed

PR #114, opened 2015-10-28
Add MPlayerX for OSX support

yshahin:
This adds the player and configures the switches to allow for URL to be passed in


xaiki:
looks good. thanks !

PR #128, opened 2015-10-29
use npm for peerflix

rom1504:
it's possible commits from peerflix master aren't quite ready to be used
peerflix is pretty well maintained


xaiki:
we'd want a specific version.

rom1504:
I put it to 0.32.1, not sure if you want to use ~0.32.1 or ^0.32.1 or  maybe 0.32.x ?

PR #131, opened 2015-10-29
Remove synchronous deletion method

### LINKED ISSUE #113: Replace fs sync functions with asynchronous versions
Fs io is very slow, it could easily lead to laggy Butter on several occasions for many people.
Replace fs sync functions with their asynchronous variants.

Just to outline worst offenders:

+ https://github.com/butterproject/butter-desktop/blob/master/src/app/app.js#L184 deleteFolder function - recursive synchronous file deleter. It should use https://www.npmjs.com/package/rimraf
+ https://github.com/butterproject/butter-desktop/blob/master/src/app/templates/settings-container.tpl#L177 recursive function - synchronous recursive directory explorer in a template file.
+ https://github.com/butterproject/butter-desktop/blob/master/src/app/lib/views/file_selector.js#L125 storeTorrent function - sync exist, unlink and write.

JEphron:
Opened to resolve #113

### DIFF
--- a/src/app/app.js
+++ b/src/app/app.js
@@ -182,29 +182,7 @@ App.addInitializer(function (options) {
 });
 
 var deleteFolder = function (path) {
-
-    if (typeof path !== 'string') {
-        return;
-    }
-
-    try {
-        var files = [];
-        if (fs.existsSync(path)) {
-            files = fs.readdirSync(path);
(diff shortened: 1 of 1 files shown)


xaiki:
awesome, feel free to fix all the things in that bug =)

PR #139, opened 2015-10-30
NSIS script update

insertc0in:
- Removed taskkill exec, bad taskkill!
- Added process detection/termination on install/uninstall
- Replaced several fixed strings with ${APP_NAME) variable
- Some indentation cleanup


xaiki:
make multiple commits !#@!@

insertc0in:
I had nothing more to commit :)

PR #141, opened 2015-10-30
NSIS script fix for windows

insertc0in:
- A "hack/fix" to allow NSIS for Windows to compile the installer when target file paths are over 260 characters long, which is the case for a lot of node modules.
- Removed a useless exclamation mark ("%s ! is the result [...]), in about.tpl
- Replaced app name instances by %s, in en.json


MrAdder:
Forgot to comment on this but WOOOOO CAN BUILD INSTALLER ON WINDOWS

PR #146, opened 2015-10-31
Added installing CMake to the install guide

Benjiko99:


hadim:
???
Butter does not use CMake.

Benjiko99:
When I ran i think 'npm install' it gave me an error that I need CMake, one of the dependencies needs it.

PR #158, opened 2015-11-02
Try to detect platform build

insertc0in:
- Instead of forcing the user to specify  -DARCH=<architecture> argument
to the makensis command, this tries to compile one or the other
depending on the build folder found.
- Changed vars names


xaiki:
you have the s/PT_/BT_/ mixed in other changes.

insertc0in:
@xaiki Where?

PR #159, opened 2015-11-02
fix #137 and add TorrentCollections

### LINKED ISSUE #137: Dropping magnet urls doesn't seem to be working for me
So I cloned the master branch and am able to run butter-desktop. However, I tried dropping a magnet url on screen and it gets stuck at the "Connecting.." screen after resolving the torrent. In console as well I just get "Loading Torrent" as a message and nothing seems to be happening after that. Any idea on how to debug this to get the magnet urls to work as they used to in PopCorn Time. Anyone facing a similar issue, any help is appreciated :)

mike1pol:
fix #137 and add Torrent Collections


src/app/lib/streamer.js
```diff
@@ -244,9 +244,7 @@
 
     var Streamer = {
         start: function (model) {
-            var torrent = model.get('torrent');
-            var torrentUrl = torrent.url ? torrent.url : torrent.magnet;
-
+            var torrentUrl = model.get('torrent');
```
xaiki: that will break archive.js

src/app/templates/torrent_collection.tpl
```diff
@@ -1,9 +1,6 @@
 <div class="torrent-collection-container">
     <div class="margintop"></div>
     <div class="content">
-        <h1>Torrent Collection is disabled in this Butter build</h1>
```
xaiki: there is a reason for this...

xaiki:
we disabled torrent collection for a reason, if you want this merged make it a module.
DO NOT MERGE.

xaiki:
can you split out torrent collection so we can merge this ?

MrAdder:
As no reply from original Author of PR I am closing this PR

PR #168, opened 2015-11-06
Don't set unexisting file as executable

### LINKED ISSUE #169: Grunt build on Linux
Running 'grunt build' on linux =>

Running "shell:setexecutable" (shell) task
chmod: cannot access ‘build/Butter/osx*/Butter.app’: No such file or directory

Made PR  #168 

Not sure it's the right way to fix the issue but it worked for me

tgroutars:
Running 'grunt build' resulted in the following error on linux:
Running shell:setexecutable (shell) task
chmod: cannot access ‘build/Butter/osx*/Butter.app’: No such file or directory

This fixes the issue by only setting the existing file as executable
Fixes #169

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -310,11 +310,14 @@ module.exports = function (grunt) {
             setexecutable: {
                 command: function () {
                     if (host.linux || host.mac) {
-                        return [
-                            'pct_rel="build/' + projectName + '"',
-                            'chmod -R +x ${pct_rel}/osx*/' + projectName + '.app || : ',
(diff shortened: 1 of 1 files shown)


xaiki:
that won't work because you could build osx on linux at somepoint. you need  to check what we're building for.

tgroutars:
What would be a better way to do it? 
Checking if the file exists?

tgroutars:
@xaiki This seems more like it, but still not sure, can you have a look?
(Sorry, I just cloned the project. I'd like to help more, I just don't know anything about it yet)

xaiki:
looks better =)

PR #187, opened 2015-11-15
fix magnet

mciparelli:
this attempts to fix https://github.com/butterproject/butter-desktop/issues/137

### DIFF
--- a/src/app/lib/streamer.js
+++ b/src/app/lib/streamer.js
@@ -244,8 +244,7 @@
 
     var Streamer = {
         start: function (model) {
-            var torrent = model.get('torrent');
-            var torrentUrl = torrent.url ? torrent.url : torrent.magnet;
+            var torrentUrl = model.get('torrent');
 
             var torrent_read = false;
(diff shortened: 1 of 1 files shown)


src/app/lib/streamer.js
```diff
@@ -244,8 +244,7 @@
 
     var Streamer = {
         start: function (model) {
-            var torrent = model.get('torrent');
-            var torrentUrl = torrent.url ? torrent.url : torrent.magnet;
+            var torrentUrl = model.get('torrent');
```
xaiki: did you test that this doesn't break archive.org support ?

MrAdder:
Can you confirm this works @mciparelli ?

src/app/lib/streamer.js
mciparelli: I haven't, sorry. I just tested drag&dropping a torrent and a magnet link and both worked.

mciparelli:
@MrAdder works with magnet and torrents for me.

PR #211, opened 2015-11-22
Fixed kat-api and nedb dependencies versions.

larruda:
Master wasn't building successfully. I did some research and found out these dependencies versions were incorrect, so fixing with this diff worked just fine. I also noticed cmake is required for building too, but it looks like there's no verification for its existence yet. I'll open/search one other issue for this.


MrAdder:
This is no longer needed

PR #224, opened 2015-12-06
Make use of i18n plural helper

evuez:
This allows to have "1 hour remaining" or "2 hours remaining" instead of "2 hour(s) remaining".
I guess every translations should be updated before this is merged though... Right?


evuez:
Anyone? It's no big thing but it's still nice to have that!

xaiki:
@insertc0in can you comment on this ?

xaiki:
looks good to me, if @insertc0in is ok, please merge.

insertc0in:
I somehow missed the notification for this commit.
I think we did this in a previous project but failed to bring over those changes into Butter.
@evuez Thanks for the commit!

PR #227, opened 2015-12-08
Update show_detail.js

iDaN5x:
this is probably the "better way"...


MrAdder:
Will test this locally here before merging :)

PR #247, opened 2016-01-10
docs : Provide SO link for npm permission problem solution

kamyarg:
Some time ago I read somewhere it is bad to install/run `npm` with `sudo` ([a similar post](https://blog.explosionpills.com/dont-use-sudo-with-npm/) )  
I had the same problem and finally [a SO answer](http://stackoverflow.com/a/24404451/1329429) solved the problem completely(I am now using `nvm` and happy again)  

I replaced the suggestion to use npm with sudo with a reference to mentioned answer.

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ The [master](https://github.com/butterproject/butter-desktop) branch which conta
 
 If you encounter trouble with the above method, you can try:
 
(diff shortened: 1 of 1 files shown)


xaiki:
with this we loose info, we still may need to use sudo. you can add your link as a suggestion but don't remove info.

PR #250, opened 2016-01-15
Vkj/patch upcoming

vankasteelj:


vankasteelj:
I don't know dafuq went wrong, I just `checkout -b` out of your branch. You forced push, didn't you ?... Anyway, here it is.

PR #282, opened 2016-02-16
if providers don't return name ignore them

ptfc503:
Signed-off-by: Niv Sardi <xaiki@evilgiggle.com>


xaiki:
i'm not sure why we'd want this !? also everything looks weird.

PR #290, opened 2016-02-22
Fix travis errors

alxhotel:
When building with travis, errors will occured because of "jshint" validation, so the build fails.

This PR just fixes those jshint erros. A great part of them it was just moving a function or variable declarations, so the functionality should be intact.

(You can checkout new travis log of this PR at: https://travis-ci.org/butterproject/butter-desktop/builds/110855410)

These are the errors that travis shows and are now fixed:
```
src/app/lib/streamer.js
    246 |    var Streamer = {
                 ^ [W003] 'Streamer' was used before it was defined.
    365 |                    var title = model.get('title');
                                 ^ [W003] 'title' was used before it was defined.
   src/app/lib/device/ext-player.js
     69 |    function getPlayerSubSwitch(loc) {
                      ^ [W003] 'getPlayerSubSwitch' was used before it was defined.
     74 |    function getPlayerFilenameSwitch(loc) {
                      ^ [W003] 'getPlayerFilenameSwitch' was used before it was defined.
     79 |    function getPlayerUrlSwitch(loc) {
                      ^ [W003] 'getPlayerUrlSwitch' was used before it was defined.
     89 |    function getPlayerSwitches(loc) {
                      ^ [W003] 'getPlayerSwitches' was used before it was defined.
     94 |    function getPlayerFS(loc) {
                      ^ [W003] 'getPlayerFS' was used before it was defined.
     99 |    var players = {
                 ^ [W003] 'players' was used before it was defined.
   src/app/lib/device/generic.js
    120 |    var _getClosestIP = function (ips, targetIp) {
                 ^ [W003] '_getClosestIP' was used before it was defined.
   src/app/lib/models/generic_collection.js
     43 |                console.log('pre---', subtitle, metadata, torrents)
                                                                            ^ [W033] Missing semicolon.
     45 |                function getDataFromProvider(torrentProvider) {
                         ^ [W082] Function declarations should not be placed in blocks. Use a function expression or move the statement to the top of the outer function.
     60 |                    console.log('pre all', promises)
                                                             ^ [W033] Missing semicolon.
     64 |                            console.log('post all', torrents, subtitles, metadatas)
                                                                                            ^ [W033] Missing semicolon.
    159 |                console.error('cached error', e)
                                                         ^ [W033] Missing semicolon.
   src/app/lib/providers/generic.js
     54 |    function installProvider(PO) {
                      ^ [W003] 'installProvider' was used before it was defined.
   src/app/lib/providers/torrent_cache.js
    128 |    var handlers = {
                 ^ [W003] 'handlers' was used before it was defined.
   src/app/lib/providers/trakttv.js
    707 |    var isValid = function (id) {
                 ^ [W003] 'isValid' was used before it was defined.
   src/app/lib/providers/watchlist.js
     40 |        function fetchWatchlist(update) {
                          ^ [W003] 'fetchWatchlist' was used before it was defined.
   src/app/app.js
    198 |    function removeCookie(cookie) {
                      ^ [W003] 'removeCookie' was used before it was defined.
    367 |    var tray = new gui.Tray({
                 ^ [W003] 'tray' was used before it was defined.
   src/app/bootstrap.js
     57 |    function loadFromPackageJSON(regex, fn) {
                      ^ [W003] 'loadFromPackageJSON' was used before it was defined.
     70 |    function loadFromNPM(name, fn) {
                      ^ [W003] 'loadFromNPM' was used before it was defined.
   src/app/httpapi.js
    747 |    function butterCallback(callback, err, result) {
                      ^ [W003] 'butterCallback' was used before it was defined.
   src/app/settings.js
    318 |            var timeout = setTimeout(fn, 5000);
                         ^ [W003] 'timeout' was used before it was defined.
    355 |        function tryNextEndpoint() {
                          ^ [W003] 'tryNextEndpoint' was used before it was defined.
   src/app/updater.js
    229 |    function extractSimple(pack, downloadPath) {
                      ^ [W003] 'extractSimple' was used before it was defined.
```

### DIFF
--- a/src/app/lib/streamer.js
+++ b/src/app/lib/streamer.js
@@ -21,7 +21,7 @@
     var watchState = function (stateModel) {
 
 
-        if (engine != null) {
+        if (engine !== null) {
 
             var swarm = engine.swarm;
             var state = 'connecting';
@@ -190,59 +190,6 @@
 
     };
 
-
-    var Preload = {
-        start: function (model) {
-
-            if (Streamer.currentTorrent && model.get('torrent') === Streamer.currentTorrent.get('torrent')) {
-                return;
-            }
-            this.currentTorrent = model;
-
-            win.debug('Preloading model:', model.get('title'));
-            var torrent_url = model.get('torrent');
-
-            readTorrent(torrent_url, function (err, torrent) {
-
-                win.debug('Preloading torrent:', torrent.name);
-                var tmpFilename = torrent.infoHash;
-                tmpFilename = tmpFilename.replace(/([^a-zA-Z0-9-_])/g, '_'); // +'-'+ (new Date()*1);
-                var tmpFile = path.join(App.settings.tmpLocation, tmpFilename);
-                subtitles = torrent.subtitle;
-
-                var torrentPeerId = crypt.pseudoRandomBytes(10).toString('hex');
-
-                win.debug('Preloading movie to %s', tmpFile);
-
-                preload_engine = peerflix(torrent_url, {
-                    connections: parseInt(Settings.connectionLimit, 10) || 100, // Max amount of peers to be connected to.
-                    dht: parseInt(Settings.dhtLimit, 10) || 50,
-                    port: 0,
-                    tmp: App.settings.tmpLocation,
-                    path: tmpFile, // we'll have a different file name for each stream also if it's same torrent in same session
-                    index: torrent.file_index,
-                    id: torrentPeerId
-                });
-
-            });
-
-
-        },
-
-        stop: function () {
-
-            if (preload_engine) {
-                if (preload_engine.server._handle) {
-                    preload_engine.server.close();
-                }
-                preload_engine.destroy();
-                win.info('Preloading stopped');
-            }
-
-            preload_engine = null;
-        }
-    };
-
     var Streamer = {
         start: function (model) {
             var torrent = model.get('torrent');
@@ -324,7 +271,7 @@
                             info: torrent,
                             subtitle: model.get('subtitle'),
                             defaultSubtitle: model.get('defaultSubtitle'),
-                            title: title,
+                            title: model.get('title'),
                             tvdb_id: model.get('tvdb_id'),
                             imdb_id: model.get('imdb_id'),
                             episode_id: model.get('episode_id'),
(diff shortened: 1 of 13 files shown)


src/app/settings.js
```diff
@@ -309,13 +312,14 @@ var AdvSettings = {
...
                         _url.hostname,
                         e.message);
-                    clearTimeout(timeout);
                     tryNextEndpoint();
                 });
+            }).setTimeout(5000, function () {
+				win.warn('[%s] Endpoint timed out',
+					_url.hostname);
+				request.abort();
+				tryNextEndpoint();
```
xaiki: this is a functional change, split it out

src/app/settings.js
```diff
@@ -280,20 +280,23 @@ var AdvSettings = {
...
-                    req.abort();
-                    tryNextEndpoint();
-                };
-            };
             var request = http.get({
                 hostname: _url.hostname
             }, function (res) {
                 res.once('data', function (body) {
-                    clearTimeout(timeout);
                     res.removeAllListeners('error');
```
xaiki: this removes the clearTimeout, it looks wrong.

xaiki:
very good work, thanks a lot, still a functional change in the middle of moving code arround, next time please split each of these chunks in their own commit so if anything breaks we can easily bissect.

src/app/lib/streamer.js
```diff
@@ -324,7 +271,7 @@
                             info: torrent,
                             subtitle: model.get('subtitle'),
                             defaultSubtitle: model.get('defaultSubtitle'),
-                            title: title,
+                            title: model.get('title'),
```
vankasteelj: This is not right, it's `title` and not `model.get('title')` for a reason

alxhotel:
That's why I said "a great part". I should have mentioned that change, though.

Instead of using a "setTimeout" outside the request to then abort it and retry again; it uses a "setTimeout" from the request object, so that you can get rid of the hassle of "clearTimeout" and variables not defined, etc.

Why it looks wrong ?

src/app/lib/streamer.js
alxhotel: Why is that ?

src/app/settings.js
alxhotel: That's why I said "a great part". I should have mentioned that change, though.

Instead of using a `setTimeout` outside the request to then abort it and retry again; it uses a `setTimeout` from the request object, so that you can get rid of the hassle of `clearTimeout` and variables not defined, etc.

Why it looks wrong ?

src/app/lib/streamer.js
vankasteelj: (I edited above)
Because 'title' is what will be used by loading.js (see: `streamInfo.set('title', torrent.title);`) to display an actual title, parsed, matched, and regexed, not just the basic title.

Your change makes no difference on providers giving a good title in the api directly, but it prevents dropped magnet/torrents from displaying the real title.

src/app/settings.js
jamesyc: I don't think it's a functional change, it's moving timeoutWrapper on line 284 and the timeout itself on line 318 and combining them. 

I'm not sure why timeoutWrapper() exists, as I don't think `request` ever changes in this context, so there's no reason for the function-in-function stuff.

jamesyc:
@alxhotel next time it'll probably be a good idea to push this commit though here (upstream) before submitting a PR to popcorntime, since it'll probably be annoying to git magic it back

src/app/lib/streamer.js
jamesyc: Rip https://github.com/popcorn-official/popcorn-desktop/commit/ae187dac12ddf00711ac3234130f6f36761f2e97

vankasteelj: I haven't noticed anything else, no. It's good work, that's sure. I just point out what I see^^

alxhotel:
@jamesyc Wow. (I just saw that commit in the popcorntime repo) I'm pretty sure I didn't commit anything to popcorntime. It's obviously my commit, but I didn't push it or anything. I guess someone there used this PR and pushed it as me. Just for the record, I'm in the legal side of this, butter's side...

vankasteelj:
@alxhotel indeed it is, PR is signed off by team-pct

jamesyc:
Someone should sign off on this PR and then revert the `title: model.get('title')` line in another commit, imo.

vankasteelj:
@alxhotel are you planning of splitting into separate commits or should I do it?

jamesyc:
@vankasteelj I'd suggest not splitting it, since that causes a git conflict and possibly a need for a rebase downstream. It's probably to accept this pull request and then just make another commit that reverts the one line about `title` afterwards.

vankasteelj:
No, it's not. As xaiki said, separate functionnal and restructuration, if it was only a `model.get('title')` misplaced, it could be ignore and fixed afterward.

Commit 1: change functions and vars without touching them, just moving around
Commit 2: modify how the timeout works

jamesyc:
The timeout change is not a functional change, though? It doesn't change the functionality, it restructures `timeoutWrapper()` into one block. 

There isn't much of a difference between the rest of moving code around and moving the timeout, from a computer science perspective. They're both just bugfixes, and code is moved in a way that can cause unintended side effects, for both cases.

I thought you were focusing on separating "bugfix/new feature" in different pull requests, which does make a lot of sense- that's why I was talking about the `title` line. That's the only feature change, I think.

jamesyc:
That's a bit overkill in this case, there's not exactly a lot to look for when you git-bisect here. If anything goes wrong, you already know where to look.

xaiki:
please do not bikeshed on this, @alxhotel do you want to do the splitting or let @vankasteelj do it ? cheers.

alxhotel:
I think it's more efficient if @vankasteelj splits it, that way he can apply whatever conventions you guys have established :)

vankasteelj:
It's done, with credits ofc. The jshint separation wasn't needed, the most important was separating the "jshint" from "setTimeout on the request object"

PR #295, opened 2016-02-24
Use opensubtitles for movies

vankasteelj:
basic implementation, it "works" but isn't fancy

### DIFF
--- a/src/app/lib/models/generic_collection.js
+++ b/src/app/lib/models/generic_collection.js
@@ -127,7 +127,7 @@
                 this.state = 'loading';
                 self.trigger('loading', self);
 
-                var subtitle = App.Providers.get('ysubs');
+                var subtitle; //TODO: var subtitle = App.Providers.get('ysubs');
                 var metadata = this.providers.metadata;
                 var torrents = this.providers.torrents;
 
--- a/src/app/lib/views/movie_detail.js
+++ b/src/app/lib/views/movie_detail.js
@@ -208,13 +208,18 @@
         },
 
         startStreaming: function () {
+            var movieInfo = {
+                type: 'movie',
+                imdbid: this.model.get('imdb_id')
+            };
             var torrent = this.model.get('torrents')[this.model.get('quality')];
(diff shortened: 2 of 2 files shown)


vankasteelj:
ofc, it uses the same system as tv, without loading subs on details, only on loading screen.

xaiki:
this should be using this.provider.subtitle, it is not, what magic black vodo are you pulling ?

vankasteelj:
I removed the subtitle provider from the top (that used to fetch subs as it was 'listing' available contents) by setting it to `undefined`, and instead made a per-content search based on IMDB, triggered inside `streamer.js`

xaiki:
we have pluggable per provider defined subtitles providers, we should use that, your changes don't touch streamer.js are you sure you've got all commited ?

vankasteelj:
I am sure. Just look for model.get('extractSubtitles') on streamer, that's where the magic happens

xaiki:
your commit can't go in like this,
 * if you want to remove subtitles from generic_collection (debatable) we should remove all subtitle code, not just turn it into dead code.
 * that overloding of the movie model seems overkill, looking at the code it looks like only passing imdb_id should work
 * the way it's done doesn't allow for different subtitles providers for different providers, and we know that is wrong, we should use the provider info to suggest a sub provider
 * i don't see a way to pass a subtitle automatically from the provider.


my suggestion to go forward is:
 - add subtitles to the config of the providers
 - load subtitles providers the same way we load providers on demand in bootstrap so we catch errors early
 - keep the subtitles loading code in generic_collection, cache them but don't wait on them (i.e. keep a promise on them)
 - remove subtitles code from streamer as it seems like the wrong place for that.

vankasteelj:
Opensubtitle can't be used as a provider: the API has a limit/ip. If you call OpenSubtitles for each item in Browser, you'll quickly hit the limit (I think it's 200 per 24h per IP, and 200 per 24h per account)

vankasteelj:
More complete work is in xaiki/providers-opts

PR #302, opened 2016-03-01
Add time-grunt to measure Grunt execution time

AdvancedCori:
Measuring the time taken for Grunt tasks to complete is quite useful insight. The package is well contained and shouldn't need any special attention.


xaiki:
thanks a lot, let's get scared and switch to gulp =)

AdvancedCori:
hehe, gulp would speed things up a lot

0tho:
Is gulp that much better and why?

AdvancedCori:
- It's a lot faster than grunt as it uses node.js stream
- Prefers writing code over weird configs
- Better quality plugins (usually)

Or you know, we could go for [webpack](https://webpack.github.io/) instead. Just food for thought.

xaiki:
patches are welcome (for gulp)

PR #336, opened 2016-03-27
Update package.json

sopex:
There is nobody that  wants smaller dimensions because the quality awful...

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "license": "GPL-3.0",
   "main": "app://host/src/app/index.html",
   "version": "0.3.8-5a",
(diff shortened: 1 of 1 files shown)


vankasteelj:
the min_height isn't the default height. It's 520 for the simple reason 1080px of height is the most used screen resolution, and that Windows is the most used platform. So 1080 - 40px of Windows menubar / 2 = 520px, hence you can put 4 applications in one screen. 

I don't see how raising the min_height by 20px would make any difference on how people use Butter, as the default height is: available screen space * 0.8

I, for example, wouldn't want those 20px to get in my whay when I multitask with 3 or 4 applications on 1 1080p screen.

PR #342, opened 2016-04-02
Fix cache folder when you load settings via npm

Bart-0110:
Without this fix the program will always use ./Temp/Butter

### DIFF
--- a/src/app/bootstrap.js
+++ b/src/app/bootstrap.js
@@ -78,6 +78,9 @@
     }
 
     App.bootstrapPromise = loadNpmSettings()
+        .then(function (){
+            Settings.tmpLocation = path.join(os.tmpDir(), Settings.projectName);
+        })
         .then(loadProviders)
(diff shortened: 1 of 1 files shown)


vankasteelj:
but there's a setting for that... the goal is to allow NOT using temp. On my netbook, I only have 1GB of free space on my builtin SSD, and I can tell the app to use my SD Card instead.

Bart-0110:
Ah okey. I forgot that you can change the temp folder.

vankasteelj:
what was the problem with temp anyway? did it set folder at different places?

Bart-0110:
For example:
My program name is 'Test'. When I open the app and go the the settings the default cache location is %temp%/Butter instead of %temp%/Test

vankasteelj:
okay, can you open an issue about that? You can PR if you want too, but this way the issue isn't lost and forgotten :)

PR #345, opened 2016-04-02
improve UI/UX with CSS props

ZakariaRidouh:


xaiki:
same comment as for the angular version, please post before/after screenshots of how this affects us.

vankasteelj:
no css in the html file please, you can set that in src/app/styl/views/misc.styl, it'll be on the top of the compiled css file

vankasteelj:
There's probably an issue in your commits, as there's no app.css file present

xaiki:
screenshots please

src/app/styl/views/main_window.styl
```diff
@@ -1,8 +1,16 @@
 html,
 body,
+    text-rendering: optimizelegibility
```
vankasteelj: this actually won't work. you're removing height/cursor/user-select/bg-color from body

src/app/styl/views/main_window.styl
```diff
@@ -1,8 +1,16 @@
...
+    -webkit-font-smoothing: subpixel-antialiased;
+    color: #FAFAFA
```
vankasteelj: this is hardcoded color. it's probably not a good idea since we work with themes for the colors

vankasteelj:
I'm failing to see any change, is it normal?

PR #354, opened 2016-04-07
Remove unneeded dependencies

ZakariaRidouh:
This helps reduce the size of the App (my Butter directories are massive)


vankasteelj:
that's a good idea, but isnt gitlab still defined in about?

xaiki:
again, string changes in the mix, aren't those modules used anywhere ? their functionalities should be removed at the same time.

src/app/templates/about.tpl
```diff
@@ -19,16 +19,16 @@
         <div class="text-about">
             <div class="full-text">
                 <%= i18n.__("%s is the result of many developers and designers putting a bunch of APIs together to make the experience of watching torrent movies as simple as possible.", Settings.projectName) %><br/>
-                <%= i18n.__("We are an open source project. We are from all over the world. We love our movies. And boy, do we love popcorn.") %>
+                <%= i18n.__("We are an open source project. We are from all over the world. We love our movies. And boy, do we love our Popcorn.") %>
```
vankasteelj: nuuuuuuuuh! you broke internationalization

xaiki:
i'm pretty sure you're breaking stuff.

```sh
git grep strike src/app/lib
src/app/lib/views/torrent_collection.js:                var strike = require('strike-api');
src/app/lib/views/torrent_collection.js:                strike.search(input, category).then(function (result) {
```
```sh
git grep gitlab src/app/lib
src/app/lib/views/issue.js:    var BT_id = 13, //ID of project (got with gitlab.projects.all). 13 is for 'butterproject/butter'
src/app/lib/views/issue.js:            if (AdvSettings.get('gitlabPassword') && AdvSettings.get('gitlabMail')) {
src/app/lib/views/issue.js:                $('#issue-email').val(AdvSettings.get('gitlabMail')),
src/app/lib/views/issue.js:                    $('#issue-pw').val(AdvSettings.get('gitlabPassword'));
src/app/lib/views/issue.js:            var gitlab = require('gitlab')({
src/app/lib/views/issue.js:            gitlab.projects.issues.list(BT_id, {
src/app/lib/views/issue.js:            var gitlab = require('gitlab')({
src/app/lib/views/issue.js:            gitlab.issues.create(
src/app/lib/views/issue.js:                    AdvSettings.set('gitlabMail', $('#issue-email').val());
src/app/lib/views/issue.js:                    AdvSettings.set('gitlabPassword', $('#issue-pw').val());
src/app/lib/views/issue.js:            var gitlab = require('gitlab')({
src/app/lib/views/issue.js:            gitlab.users.session(email, password, function (response) {
```

src/app/templates/about.tpl
kiefferbp: I prefer the old text. I'm not sure how this line is supposed to be an improvement.

PR #395, opened 2016-05-15
Bollywood filter

krackjax:
Would like for a way to find Indian movies by adding a Bollywood filter as a Genre


xaiki:
this pr is clearly not for us.

PR #406, opened 2016-05-31
Xaiki/audio dropdown

xaiki:
this opens the posibility to have providers defining audio as well as subs, it's WIP as the provider API will require at least a new method (selectStream), that will actually allow us to move some logic that should not be in generic code into the provider (like quality selection)

flow will be something like:
provider (declare selectable) -> generic code (presents selectable) -> provider (select stream)

but it's already a nice refactor of the sub code as it is.

### DIFF
--- a/src/app/app.js
+++ b/src/app/app.js
@@ -1,4 +1,6 @@
 // Special Debug Console Calls!
+var localStorage = window.localStorage || {};
+
 win.log = console.log.bind(console);
 win.debug = function () {
     var params = Array.prototype.slice.call(arguments, 1);
--- a/src/app/language.js
+++ b/src/app/language.js
@@ -23,6 +23,12 @@ var setLanguage = function (preferredLanguage) {
     });
 };
 
+
+App.Localization.nativeName = function (lang) {
+    var codes = App.Localization.langcodes
+    return codes[lang]?codes[lang].nativeName:'No native Name for: '+lang;
+}
+
 App.Localization.detectLocale = function () {
     // The full OS language (with localization, like 'en-uk')
     var pureLanguage = navigator.language.toLowerCase();
--- a/src/app/language.js
+++ b/src/app/language.js
@@ -27,7 +27,7 @@ var setLanguage = function (preferredLanguage) {
 App.Localization.nativeName = function (lang) {
     var codes = App.Localization.langcodes;
     return codes[lang]?codes[lang].nativeName:lang;
-}
+};
 
 App.Localization.detectLocale = function () {
     // The full OS language (with localization, like 'en-uk')
--- a/src/app/language.js
+++ b/src/app/language.js
@@ -25,8 +25,8 @@ var setLanguage = function (preferredLanguage) {
 
 
 App.Localization.nativeName = function (lang) {
-    var codes = App.Localization.langcodes
-    return codes[lang]?codes[lang].nativeName:'No native Name for: '+lang;
+    var codes = App.Localization.langcodes;
+    return codes[lang]?codes[lang].nativeName:lang;
 }
 
 App.Localization.detectLocale = function () {
--- a/src/app/lib/views/lang_dropdown.js
+++ b/src/app/lib/views/lang_dropdown.js
@@ -13,18 +13,18 @@
 
         initialize: function () {
             this.type = this.model.get('type');
-            this.handler = this.model.get('handler')
-            this.selected = this.model.get('selected') || 'None'
+            this.handler = this.model.get('handler');
+            this.selected = this.model.get('selected') || 'None';
         },
         onShow: function () {
-            console.log ('show', this)
+            console.log ('show', this);
         },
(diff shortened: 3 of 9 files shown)


src/app/app.js
```diff
@@ -1,4 +1,6 @@
 // Special Debug Console Calls!
+var localStorage = window.localStorage || {};
```
vankasteelj: it's already defined in nwjs, isn't it?

src/app/language.js
```diff
@@ -23,6 +23,12 @@ var setLanguage = function (preferredLanguage) {
     });
 };
 
+
+App.Localization.nativeName = function (lang) {
+    var codes = App.Localization.langcodes
+    return codes[lang]?codes[lang].nativeName:'No native Name for: '+lang;
+}
```
vankasteelj: This doesn't pass jshint tests (missing semicolon and spaces). I also wonder if the condition are good, I mean, it's not because you have codes[lang] that you automatically have codes[lang].nativeName. I'm also wondering if you shouldn't try to solve the issues elsewhere instead of sending back a "error" message. How about:
```
App.Localization.nativeName = function (lang) {
    var codes = App.Localization.langcodes;
    return codes[lang] && codes[lang].nativeName ? codes[lang].nativeName : null;
};```

File: src/app/lib/views/lang_dropdown.js [hunk 3]
```diff
@@ -0,0 +1,53 @@
...
+
+        initialize: function () {
+            this.type = this.model.get('type');
+            this.handler = this.model.get('handler')
+            this.selected = this.model.get('selected') || 'None'
+        },
+        onShow: function () {
+            console.log ('show', this)
+        },
+        toggleDropdown: function (e) {
```
vankasteelj: this looks wrong. `values` is never used and you use find (slow) a tons lot.

vankasteelj:
apart from jshint and some comments, it looks ok to me

src/app/app.js
xaiki: it was failing on my machine and throwing lots of errors everytime i'd move the window.

src/app/language.js
xaiki: i had that thought, i think instead of null i should return `lang` that would make more sense... (it's used i templates where we really don't want to do err handeling)

src/app/lib/views/lang_dropdown.js
xaiki: i used find because $() was not scoping correctly and hitting the other components... oh the joys of backbone...

xaiki: yeah that values call needs to go.

xaiki:
@vankasteelj fixed jshinting and your comments.

src/app/language.js
xaiki: made it return lang

src/app/lib/views/lang_dropdown.js
xaiki: values gone, the find, i can't remove because of scoping, you have a better solution ?

src/app/app.js
vankasteelj: If you're 100% sure it won't destroy localStorage for the entire session if it fails to init, I don't see a problem, but from the outside, it looks like (if there's a fail) you replace the global variable by a volatile object that won't be saved on exit, hence that won't be available for next session. So maybe you should add in the try/catch of win.close() a `window.localStorage = localStorage` to make sure the modifications to localStorage are saved if localStorage is available AFTER the app started. Dunno if I make sense ? :p

xaiki: i think we should use localforage that we know will not fail and will handle all that, you in ?

vankasteelj: If it's a replacement that works, why not! My pov is mostly that having a "quick and easy" db-like (like localStorage) for little things, it's a plus (over nedb, for instance). So yeah.

xaiki: do we know why though ?

vankasteelj: nope, but it does it on nwjs 0.18 too, so it's either an old bug, or something else

xaiki: haha, ok, i'll push that change, that branch needs more work from what we did on the hackathon anyway.

xaiki:
so, @vankasteelj can we merge this ?

src/app/app.js
vankasteelj: Before merging, how about a app.js: win.on('close') :

```
try {
  delCache();
  saveLocalStorage();
} catch (e) {
  win.close(true);
}
```

and app.js:

```
var saveLocalStorage = function () {
  if (window.localStorage && localStorage) {
  if (JSON.stringify(localStorage) !== JSON.stringify(window.localStorage)) {
    for (entry in localStorage) {
      window.localStorage[entry] = localStorage[entry];
    }
  }
};
```

xaiki: add it, that commit is about 3 months old, i'd rather remove the localstorage part than introduce more stuff.

Issue #418, opened 2016-07-19
TypeError: handler.state is not a function on build


fernandomora:
When I follow instructions to build and run the project I receive the following error:
``` 
$ gulp build
(node:50631) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.
(node:50631) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.
[00:40:14] Using gulpfile ~/git/butter-desktop/gulpfile.js
[00:40:14] Starting 'build'...
[00:40:14] Starting 'injectgit'...
Branch: master
Commit: e3da5178
[00:40:14] Finished 'injectgit' after 19 ms
[00:40:14] Starting 'css'...
Stylus files compiled in ~/git/butter-desktop/src/app/themes/
[00:40:19] Finished 'css' after 4.86 s
[00:40:19] Starting 'nwjs'...
[00:40:53] 'nwjs' errored after 34 s
[00:40:53] TypeError: handler.state is not a function
    at Object.inspect (~/git/butter-desktop/node_modules/when/lib/state.js:28:23)
    at formatValue (util.js:348:21)
    at formatProperty (util.js:777:15)
    at util.js:637:12
    at Array.map (native)
    at formatObject (util.js:636:15)
    at formatValue (util.js:575:16)
    at formatProperty (util.js:777:15)
    at util.js:637:12
    at Array.map (native)
    at formatObject (util.js:636:15)
    at formatValue (util.js:575:16)
    at formatProperty (util.js:777:15)
    at util.js:637:12
    at Array.map (native)
    at formatObject (util.js:636:15)
[00:40:53] 'build' errored after 39 s
[00:40:53] TypeError in plugin 'run-sequence(nwjs)'
Message:
    handler.state is not a function
Stack:
TypeError: handler.state is not a function
    at Object.inspect (~/git/butter-desktop/node_modules/when/lib/state.js:28:23)
    at formatValue (util.js:348:21)
    at formatProperty (util.js:777:15)
    at util.js:637:12
    at Array.map (native)
    at formatObject (util.js:636:15)
    at formatValue (util.js:575:16)
    at formatProperty (util.js:777:15)
    at util.js:637:12
    at Array.map (native)
    at formatObject (util.js:636:15)
    at formatValue (util.js:575:16)
    at formatProperty (util.js:777:15)
    at util.js:637:12
    at Array.map (native)
[...]

PR #420, opened 2016-07-26
Fixed TypeScript Error: handle.state is not a function.

mjandrevska:
Commented the donwloadURL line from gulpfile.js, which was causing the TypeScript Error introduced in Issue #418.
Now the application can be built successfully.


xaiki:
hello, thanks for your patch,
that said i think this is not the right solution as we need a custom built nw binary, @vankasteelj can you look at this ?

vankasteelj:
mh, yeah

PR #433, opened 2016-09-10
fix nw prebuit download url

adelnizamutdinov:
Problem: 400 when trying to download prebuilt nw from `https://raw.githubusercontent.com/butterproject/nwjs-prebuilt/master/`

Solution: change it to `http://dl.nwjs.io/`


vankasteelj:
that is not a solution, we use codecs that aren't shipped with nwjs

PR #436, opened 2016-09-20
First webtorrent implementation

VitorVRS:
- Added webtorrent to dependencies
- Required wetorrent for app
- Added specific webtorrent stream info (WIP)
- Rewrited streamer.js to use WebTorrent engine (WIP)


package.json
```diff
@@ -76,7 +76,8 @@
         "trakt.tv-ondeck": "0.x",
         "underscore": "1.x.x",
         "upnp-mediarenderer-client": "1.x.x",
-        "urijs": "1.x.x"
+        "urijs": "1.x.x",
+        "webtorrent": "^0.94.4"
```
xaiki: probably need to remove peerflix ?

src/app/global.js
```diff
@@ -51,5 +51,7 @@ var
     // Torrent engines
     peerflix = require('peerflix'),
```
xaiki: remove peerflix ?

src/app/global.js
```diff
@@ -51,5 +51,7 @@ var
...
 
+    WebTorrent = require('webtorrent');
+
```
xaiki: so you add webtorrent but don't use  it ?

src/app/index.html
```diff
@@ -84,6 +84,7 @@
 	<script src="lib/models/generic_collection.js"></script>
 	<script src="lib/models/movie_collection.js"></script>
 	<script src="lib/models/stream_info.js"></script>
+	<script src="lib/models/webtorrent_stream_info.js"></script>
```
xaiki: shouldn't we remove the other one ?

src/app/lib/models/webtorrent_stream_info.js
```diff
@@ -0,0 +1,97 @@
...
+            this.on('change:size', function () {
+                this.set('sizeFormatted', Common.fileSize(this.get('size')));
+            }.bind(this));
+
+            this.set('size', 0);
+        },
+
+        selectFile: function () {
+
+            var wtTorrent = this.get('wtTorrent'),
```
xaiki: what is wtTorrent ?

src/app/lib/streamer.js
```diff
@@ -1,501 +1,491 @@
-(function (App) {
+(function (App, WebTorrent) {
```
xaiki: this looks wrong, why is it needed ?

package.json
VitorVRS: Yes, but there are some sources which are using peerflix that should be reviewed before.

src/app/global.js
VitorVRS: Added those requires following the project pattern, the same way as peerflix definition above

src/app/lib/streamer.js
VitorVRS: Linter was accusing that WebTorrent wasn't defined on the scope that i used, maybe its wrong.

VitorVRS:
Just to be clear, this is an prototype and should be refactored.

src/app/lib/models/webtorrent_stream_info.js
VitorVRS: Its Torrent object from WebTorrent, which i used to get info about upload/download.

src/app/index.html
VitorVRS: webtorrent_stream_info.js its not completed, since i couldn't get the same info that peerflix supply.
I need help in this one.

vankasteelj:
At first glance it looks great ! A question : how do you check if it's a magnet, a torrent or an url ?

VitorVRS:
There are two "types" of torrent: torrent from butter api and torrent inputed by user.
If it's torrent from api, i use Common.matchTorrent to get metadatas and start streaming.
If it's from user's input, i use read-torrent to get torrent info and open the file selector.
https://github.com/VitorVRS/popcorn-desktop/blob/dc93dff892ae089da78825a14ecf96c22d708e2b/src/app/lib/streamer.js#L96

This is the same behaviour than before.

vankasteelj:
Before, we had :

```
            // HACK(xaiki): we need to go through parse torrent
            // if we have a torrent and not an http source, this
            // is fragile as shit.
            if (typeof (torrentUrl) === 'string' && torrentUrl.substring(0, 7) === 'http://' && !torrentUrl.match('\\.torrent') && !torrentUrl.match('\\.php?')) {
                return Streamer.startStream(model, torrentUrl, stateModel);
            } else if (!torrent_read) {
                readTorrent(torrentUrl, doTorrent);
            } else {
                doTorrent(null, model.get('torrent'));
}
```

src/app/lib/models/webtorrent_stream_info.js
xaiki: why not torrent ?

src/app/index.html
xaiki: we should design this PR to nuke peerflix, can you be more specific about what you're missing ?

src/app/global.js
xaiki: it looks like you could just import webtorrent in streamer.js doesn't it, @vankasteelj is there anything against just doing that ?

package.json
xaiki: yeah, we won't merge before properly reviewing, please make that PR peerflix removing =)

xaiki:
i don't feel it needs much refactor, just a few things to iron out and we're ready to merge this, FINALLY after 2 years of looking into it =)

src/app/lib/models/webtorrent_stream_info.js
xaiki: or `info` if torrent is already used...

ChrisAlderson:
> 2 years of looking into it =)

Open-source is great... eventually.

VitorVRS:
@vankasteelj if you are refering to this code:
```javascript
if (typeof (torrentUrl) === 'string' && torrentUrl.substring(0, 7) === 'http://' && !torrentUrl.match('\\.torrent') && !torrentUrl.match('\\.php?')) {
    return Streamer.startStream(model, torrentUrl, stateModel);
}
```

i removed because it doesn't make sense to me this treatment. 
If you explain to me the problem which originated that hack i could re-implement it.

src/app/index.html
VitorVRS: stream_info.js, for instance, shows information about how many peers are and how many peers are active, but webtorrent shows number of peers (??  https://github.com/feross/webtorrent/blob/master/docs/api.md#torrentnumpeers).

I couldn't show the same information  which stream_info.js show to the user.
This is why i created another file, for comparison.
Maybe i'll need help on this file, on what information i really need to show.

Sorry if couldn't be clear enough.

xaiki:
@VitorVRS it's due to how broken our streamer has grown to be, the streamer actually resolves the torrent, opens it, finds the file, and then decides what to stream, if we get a direct `http://` url, that is not a torrent, we just ask it to stream. the code that @vankasteelj pasted explains it better: 

```js
            // HACK(xaiki): we need to go through parse torrent
            // if we have a torrent and not an http source, this
            // is fragile as shit.
            if (typeof (torrentUrl) === 'string' && torrentUrl.substring(0, 7) === 'http://' && !torrentUrl.match('\\.torrent') && !torrentUrl.match('\\.php?')) {
                return Streamer.startStream(model, torrentUrl, stateModel);
            } else if (!torrent_read) {
                readTorrent(torrentUrl, doTorrent);
            } else {
                doTorrent(null, model.get('torrent'));
}
```

src/app/lib/streamer.js
VitorVRS: here is why: https://github.com/butterproject/butter-desktop/blob/master/.jshintrc#L80
i didn't put webtorrent on .jshintrc

xaiki:
i can confirm that http streaming is broken on master, so let's get this merged as well as we can and then move to fix the rest of the streamer.

xaiki:
so i think we need:
[ ] remove peerflix
[ ] basic stats

vankasteelj:
it doesn't work at all on my part, it won't resolve a magnet nor start downloading a torrent

src/app/lib/streamer.js
xaiki: it looks like you should just import webtorrent in that file, or pass it via App. we have enough global variables we don't want more =/

src/app/lib/models/webtorrent_stream_info.js
```diff
@@ -0,0 +1,97 @@
...
+            if (isNaN(downloadTimeLeft) || downloadTimeLeft < 0) {
+                downloadTimeLeft = 0;
+            } else if (!isFinite(downloadTimeLeft)) { // infinite
+                downloadTimeLeft = undefined;
+            }
+
+            this.set('pieces', 0);
+            this.set('downloaded', downloaded);
+            this.set('active_peers', wtTorrent.numPeers);
+            this.set('total_peers', wtTorrent.numPeers);
```
VitorVRS: @xaiki Here i have these two peers information "active_peers" and "total_peers", but webtorrent gives us only "numPeers"

xaiki: i think it's ok for now maybe @guerrerocarlos can help us get better data ?

vzamanillo: Interesting torrent properties:

- _[numConns](https://github.com/feross/webtorrent/blob/master/lib/torrent.js#L200) connected peers?

- _[numQueued](https://github.com/feross/webtorrent/blob/master/lib/torrent.js#L194) peers waiting?

src/app/lib/models/stream_info.js
```diff
@@ -76,8 +77,8 @@
 
             this.set('pieces', 0);
             this.set('downloaded', downloaded);
-            this.set('active_peers', info.numPeers);
-            this.set('total_peers', info.numPeers);
+            this.set('active_peers', torrent.numPeers);
```
vzamanillo: We can use the property _[numConns](https://github.com/feross/webtorrent/blob/master/lib/torrent.js#L200) to get the active peers here.
```javascript
this.set('active_peers', torrent._numConns);`
```

and add an extra info like the queued clients with the _[numQueued](https://github.com/feross/webtorrent/blob/master/lib/torrent.js#L194) property,

```javascript
this.set('queued_peers', torrent._numQueued);
```

src/app/lib/views/file_selector.js
```diff
@@ -74,10 +74,9 @@
             var torrent = _this.model.get('torrent');
             var file = parseInt($(e.currentTarget).attr('data-file'));
             var actualIndex = parseInt($(e.currentTarget).attr('data-index'));
-            torrent.name = torrent.files[file].name;
 
             var torrentStart = new Backbone.Model({
-                torrent: torrent,
+                torrent: torrent.magnetURI,
```
xaiki: i'm afraid this may break some usecases, like pure HTTP streaming

VitorVRS: Do you have this use cases? I could test them.

xaiki: paste a link from http i.e. put a file on a webserver and throw the URL on Butter

VitorVRS: Butter will show the file selector for an http resource? Doesn't seem to make sense for me, but i'll test this cases too. Tks

VitorVRS:
Closing due to this: https://github.com/butterproject/butter-desktop/pull/475

PR #437, opened 2016-09-22
Improved AirPlay device compatibility

SteveJobzniak:
- Replace the awful "airplay-js" with @watson's modern libraries.
- All existing features of the airplay device module have been ported to @watson's libraries.


watson:
If you're interested in how I've added all the airplay features (start/stop/scrubbing/etc) to [webtorrent-desktop](https://github.com/feross/webtorrent-desktop), check out this snippet:

https://github.com/feross/webtorrent-desktop/blob/6541291e0dc666c8a8ae86915517665e11338c40/src/renderer/lib/cast.js#L178-L267

vankasteelj:
with the airplay commits backported from pt, can we close this one?

xaiki:
this uses bonjour and airplay-protocol, i'm not sure it works best with what we merged...

SteveJobzniak:
@xaiki @vankasteelj Sorry about my long absence. Team Popcorn Time's solution was based on the library I found (they referenced me in their pull request). They use "airplayer" which is the meta-package for airplay-protocol and bonjour, which hides a lot of the raw power of the underlying airplay-protocol...

It's good that you've finally changed library. However... the changes I made here were much more than just swapping the library. I added more robust discovery and tracking of Airplay devices, now hashed by host:port combination to avoid the risk of seeing duplicate services if a computer announces itself twice with two different airplay names (ie if you rename the AppleTV device).

I also noticed that the code you merged no longer un-registers AirPlay devices when they go offline.

It's a big "what the hell?" moment to see such a poor pull request merged instead of this. ;-)

src/app/lib/device/airplay.js
```diff
@@ -1,53 +1,80 @@
...
+
+    var bonjourBrowser = new Bonjour(),
         collection = App.Device.Collection;
 
-    var makeID = function (baseID) {
-        return 'airplay-' + baseID.replace(':', '');
+    var makeID = function (bonjourService) {
+        // strong ID since there can only be 1 bonjour service per host+port
+        var uniqueDeviceId = bonjourService.host + '_' + bonjourService.port;
+        return 'airplay-' + Common.md5(uniqueDeviceId);
```
xaiki: better keep the full string, there is no use of having the md5 and it makes it harder to debug

src/app/lib/device/airplay.js
```diff
@@ -1,53 +1,80 @@
...
         },
+
         play: function (streamModel) {
-            var url = streamModel.attributes.src;
-            this.device.play(url);
+            var self = this;
+            var url = streamModel.get('src');
+            this.device.player.play(url, function (err, res, body) {
+                if (err) {
+                    // FIXME: do something on failure to play
```
xaiki: yeah, really call stop()

src/app/lib/device/airplay.js
team-pct: @xaiki  it is to avoid duplicate it is good idea

team-pct:
+1  @SteveJobzniak I am making things work until get a more reliable solution like yours , but as you were gone i had to PR something that works I hope you are not MAD 🎱

team-pct:
@SteveJobzniak Can you send a new PR with those changes to avoid conflicts 
indeed it is better work than mine ^^

src/app/lib/device/airplay.js
SteveJobzniak: @xaiki Are you sure? I followed the Common.md5() conventions of the chromecast lib, which gives nice looking IDs: https://github.com/SteveJobzniak/butter-desktop/blob/5b0622661d98a782dd4aa9a447b251253d84441e/src/app/lib/device/chromecast.js#L14

I don't think we need the internal device ID to be "airplay-AppleTV.local:4834" or whatever, since we never look at that. If there's a problem with a device (for debugging), we've got all of that info available to us in the other fields of the AirPlay device object. The exampleDevice.service.name and exampleDevice.service.port fields are always available for the full details.

src/app/lib/device/airplay.js
SteveJobzniak: @xaiki this.stop() would send the airplay command to tell the AirPlay receiver to stop, and yeah that's a good basic start.

For better handling, I think we'd want something like an error message in the Butter interface, or to stop overall playback in the Butter interface.

But I am new to this project and don't know how to do that, so I'll leave those more advanced error handlings to someone else.

I'll add this.stop() for now.

SteveJobzniak:
@team-pct Haha I am only a little bit mad. ;-) No worries. We'll get this sorted out and enjoy better AirPlay together!

I'm going to bed but I'll take care of this when I wake up.

SteveJobzniak:
Okay I am nearly done rebasing but I need to ask something:

What the holy f-ck is this commit doing? @vankasteelj 

https://github.com/butterproject/butter-desktop/commit/ca627983e183bd22746d74d80dbcc72efbfdd575

That commit has no description and no commit message!

1. The network-address library is made for retrieving the ipv4 or ipv6 address of the computer that Butter Desktop is running on.
2. The commit uses the "get ipv4 address of localhost" command: `netw()`
3. It sets `this.attributes.address` (the "address" attribute of the target Device) to `netw()`: https://github.com/butterproject/butter-desktop/commit/ca627983e183bd22746d74d80dbcc72efbfdd575#diff-9eda59a09139f5bdcadea6acb49424daR23.
4. In other words... That commit did the following: Set the "IP address" of the device we're streaming to, to the IP of our local Butter Desktop.

No other device libraries do that. All other device libraries set this.attributes.address to the IP of the TARGET device: https://github.com/butterproject/butter-desktop/tree/ca627983e183bd22746d74d80dbcc72efbfdd575/src/app/lib/device

I had to stop my rewriting here because of this weird commit. But when I wake up, I am going to guess that the entire commit by @vankasteelj was wrong and I'll revert that code?

team-pct:
airplayer return a host like atv.local and our IP picker will get a wrong SRC URL if you have ipv6 on your network

SteveJobzniak:
@team-pct It's not an old commit. Here is the Master branch history for the airplay code:

https://github.com/butterproject/butter-desktop/commits/master/src/app/lib/device/airplay.js

@vankasteelj did the weird change 5 days ago. Long after your code was merged. Here:

https://github.com/butterproject/butter-desktop/commit/ca627983e183bd22746d74d80dbcc72efbfdd575

And like I showed you (in the edited comment I posted earlier), I did this:

`grep -Ri "address" .` (a case insensitive search for the word "address" in the entire Butter Project source code).

Answer: ZERO code ever reads "device.attributes.address". It's only set in the per-device libraries as a "nice thing to have" (I guess) but it's never read anywhere.

So I thought "okay, what about attributes? maybe something gets the attributes and sends them to another external library whose code I am not searching in".

`grep -Ri "attributes" .`

And the answer: No code ever looks at the playback device's "attributes". There's some code that looks at the ".attributes.id". But there is no code anywhere that gets attributes and sends it somewhere else that may look at host.

So the answer is: @vankasteelj I will revert all of the changes from https://github.com/butterproject/butter-desktop/commit/ca627983e183bd22746d74d80dbcc72efbfdd575. Except the console.error() logging.

SteveJobzniak:
Furthermore, here are some better ways to fix the streaming URL:

- Option 1: (outside airplay.js) if deviceType == AirPlay, generate an ipv4 streaming URL to give to the device's play() function.
- Option 2 (inside airplay.js): Get the local machine's ipv4 IP, and rewrite the url to replace the host part with an IPv4 address.
- Option 3 (outside airplay.js): Make it possible for device libraries to specify what type of URL their play() method wants, and make Butter Desktop give the library the proper type.

vankasteelj:
Nobody here even has an airplay tv. I couldnt care less about apple. I've never written a single line of code regarding that (they all can go fuck themselves with their overpriced hardware and locked system). If I touched anything about airplay, it's that it was in a backport of some forks here and there. So : nope. Leave me all out of your crap.

SteveJobzniak:
@vankasteelj Hehe alright. It would have helped if you had referenced the original guy's pull request so we could know what those code changes were meant to do. I had to waste an hour researching all areas of Butter Desktop to make sure the code was as bad as it looked. And it was as bad as it looked. I've taken it out.

I'm now putting final touches on the updated version of this pull request. I've rebased the history so that it will merge cleanly.

xaiki:
awesome thanks @SteveJobzniak looking forward to look at your code !

SteveJobzniak:
@xaiki It's all done except that I am also adding scrub/forward/reverse etc support now :) Soooo not long left.

SteveJobzniak:
@xaiki I have now pushed the rebased branch with updated code. Everything is cleaned up. I'll now start working on the scrub(). Just finished researching how Apple does scrubbing.

src/app/lib/device/airplay.js
```diff
@@ -1,56 +1,119 @@
...
+        _makeID: makeID,
+
         initialize: function (attrs) {
             this.device = attrs.device;
-            this.attributes.id = this.makeID(this.device.host);
-            this.attributes.name = this.device.name || this.device.serverInfo.model;
-            this.attributes.address = netw();
+            this.attributes.id = this._makeID(this.device.service);
+            this.attributes.name = this.device.service.name;
+            this.attributes.address = this.device.service.host; // receiver's hostname, ie "AppleTV.local"
```
SteveJobzniak: I added this to prevent anyone from ever setting this to the local Butter Desktop IP again. ;p

src/app/lib/device/airplay.js
```diff
@@ -1,56 +1,124 @@
...
```
SteveJobzniak: I added this comment to prevent anyone from ever setting this to the local Butter Desktop IP again. Whoever stupidly changed that (without leaving an explanation) last time wasted at least an hour of my life. ;p

src/app/lib/device/airplay.js
```diff
@@ -1,56 +1,124 @@
...
+
+    var bonjourBrowser = new Bonjour(),
         collection = App.Device.Collection;
 
-    var makeID = function (baseID) {
-        return 'airplay-' + baseID.replace('.', '');
+    var makeID = function (bonjourService) {
+        // strong de-duplication ID since there can only be 1 bonjour service per host+port
+        var uniqueDeviceId = bonjourService.host + '_' + bonjourService.port;
+        return 'airplay-' + Common.md5(uniqueDeviceId);
```
xaiki: please, remove that md5, keep the full string, it's as deduplicating and human readable.

src/app/lib/device/airplay.js
xaiki: will that work on all platforms ? like the ones that don't support mdns ?

src/app/lib/device/airplay.js
SteveJobzniak: @xaiki Just explain why Chromecast uses md5?

https://github.com/butterproject/butter-desktop/blob/master/src/app/lib/device/chromecast.js#L14

We do not need the internal de-duplication ID to be human readable.

Imagine this: We have a device that is bugged. We want to debug it. Do we look at device.id? No. We look at device.service.name and device.service.host and device.service port.

I think we should keep Common.md5(). It's a nicer variable name. And all debug info is available to us anyway.

src/app/lib/device/airplay.js
SteveJobzniak: @xaiki "service" is an instance of the bonjour-protocol by @watson. It implements mDNS == multi-cast DNS == Bonjour in a cross-platform way in pure javascript.

As for the address, if you are on Windows it seems his library chooses whatever hostname Windows prefers:

https://github.com/watson/bonjour/blob/master/lib/service.js#L23

I will clarify the code comment to clarify this OS difference.

src/app/lib/device/airplay.js
xaiki: chromecast is probably broken too. there is no point in using md5 there.
add to that device id all you need to make it deduplicatable and debuggeable, but keep it a human readable string, because shady numbers that have no justification make no sense.

src/app/lib/device/airplay.js
SteveJobzniak: Okay I have found the answer:
https://github.com/watson/bonjour#servicehost

host == whatever hostname the device was found as ON YOUR NETWORK, so on Mac it would be "AppleTV.local" and Windows it may be "192.168.0.23".

fqdn == The full mDNS service location descriptor ie "Foo Bar._http._tcp.local".

So yes this code with "host" works on Windows. I will clarify the comment.

xaiki: cool, thanks for that research !

src/app/lib/device/airplay.js
xaiki: (and they bite us later)

SteveJobzniak: Also: Many of the device libraries to string replacement on the device IDs to make them "safe"? Like this:

https://github.com/butterproject/butter-desktop/blob/master/src/app/lib/device/dlna.js#L9

By doing Common.md5 we don't need to clean up the "host/ip_port" string.

SteveJobzniak: @xaiki Alright, makes sense. I'll remove the md5().

SteveJobzniak:
@xaiki: I am working on the scrub code right now. Want to wait with the merge until that is done? I don't expect much work to remain on scrubbing but I need to know something:

What's the list of scrub-methods that a Device must implement?

I am looking at chromecast.js and it has forward(), backward(), seekPercentage(), seekRelative(). Should I do all of those?

SteveJobzniak:
@xaiki: I am working on the scrub code right now. Want to wait with the merge until that is done? I don't expect much work to remain on scrubbing (it will be done this evening) but I need to know something:

What's the list of scrub-methods that a Device must implement?

I am looking at chromecast.js and it has forward(), backward(), seekPercentage(), seekRelative(). Should I do all of those? Less? More?

Now I am looking at DLNA and it has even more methods:

https://github.com/butterproject/butter-desktop/blob/master/src/app/lib/device/dlna.js

seek(), seekTo() etc...

I am sure some of these are internal ideas invented by these device wrappers...

So I just need a full list of what BUTTER expects me to provide.

xaiki:
for now let's only do things in airplay.js, butter is not the only app running on the desktop, so it has to do the same checks as need to be done internally.

xaiki:
yes... we need to document devices better... @vankasteelj is rewritting part of the player and streamer code, once he's done i'm going to do a documentation pass.

SteveJobzniak:
@xaiki Okay, well I just need a list of the names and parameters of the scrub/position functions Butter wants my device to implement, so that people can move the playback position in butter and see the position change on their tv.

SteveJobzniak:
@xaiki Alternatively, I can push right now with the final change that also has my _scrub(seconds) internal function, and add a "// TODO: implement all Butter seek/position functions via the internal _scrub() function" for some other contributor to take care of. And leave that the scrubbing implementation to someone else after the player code has been finalized.

xaiki:
it looks like these are implemented in player.js now: ```
        stop: function() {
            this.player.stop();
        },

        pause: function() {
            this.player.pause();
        },

        forward: function() {
            this.player.seek(30);
        },

        backward: function() {
            this.player.seek(-30);
        },

        seek: function(seconds) {
            console.log('DLNA: seek %s', seconds);
            this.get('player').seek(seconds, function(err, status) {
                if (err) {
                    console.error('DLNA.seek:Error', err);
                }
            });
        },```

SteveJobzniak:
@xaiki Thanks a lot for that!

The AirPlay protocol only supports seeking to an exact number of seconds. So I cannot implement "seek +30" and "seek -30" without knowing something else: The current butter player's (desktop) position in seconds, so that I can add or remove 30 to that number.

Can you tell me where to find that too? Then I'll get this done in a moment.

SteveJobzniak:
@xaiki Wait a minute, you took that from DLNA.js ;) And it seems to be incomplete. Here's from generic.js instead:
```javascript
        forward: function () {
            self.selected.forward();
        },
        backward: function () {
            self.selected.backward();
        },
        seek: function (seconds) {
            self.selected.seekBy(seconds);
        },
        seekTo: function (newCurrentTime) {
            self.selected.seekTo(newCurrentTime);
        },
        seekPercentage: function (percentage) {
            self.selected.seekPercentage(percentage);
        },
```

SteveJobzniak:
Ahh forget it, I will push it with my internal _scrub function and then when @vankasteelj is done finalizing the player scrub/seek interface someone can easily implement all of those actions via my _scrub().

Give me 10 minutes and this will all be done! :-)

xaiki:
`this.player._status.duration ` look at the dlna.js file, it's the most complete i think.

SteveJobzniak:
@xaiki Nope, unfortunately this.player is the DLNA player itself. So that code is just looking at its own DLNA player's current position and seeking relative to that. (Which I could also do for AirPlay but that makes no sense; it's better to seek to whatever position the Butter Desktop player itself seeked to, in case the AppleTV is out of sync ie after you used the AppleTV remote to do manual seeking).

Anyway, I've pushed the final code now, and scrubbing can be added by someone who wants to attempt to figure out what Butter Project expects. :)

xaiki:
ok, so it works in `device/generic.js`
by listening to all this: ```
            App.vent.on('device:list', this.list);
            App.vent.on('device:pause', this.pause);
            App.vent.on('device:unpause', this.unpause);
            App.vent.on('device:stop', this.stop);
            App.vent.on('device:forward', this.forward);
            App.vent.on('device:backward', this.backward);
            App.vent.on('device:seek', this.seek);
            App.vent.on('device:seekTo', this.seekTo);
            App.vent.on('device:seekPercentage', this.seekPercentage);
            App.vent.on('device:status:update', this.updateStatus);
```

**Comment by xaiki** [comment]
so as you're the only player, you're going to have to get the current time from airplay itself.

**Comment by SteveJobzniak** [comment]
@xaiki Oh yeah, when you AirPlay, butter desktop doesn't have any Playback on the desktop itself. I forgot about that and thought it played on the desktop too.

I have no more time for this. The original airplay.js didn't even have pause and unpause, and didn't have seek at all. Another contributor can pick up from this clean code and extend it with the seeking in the future :)

**Comment by SteveJobzniak** [comment]
@xaiki Oh yeah, when you AirPlay, butter desktop doesn't have any Playback on the desktop itself. I forgot about that and thought it played on the desktop too.

I cannot test actual AirPlay streaming coming from Butter, since Butter's HTTP server (for the media file serving) has no support for the AppleTV4, so AppleTV4 fails to connect and start streaming the URL it's given by airplay.js. (That's a separate issue in Butter's HTTP server and has _nothing_ to do with the airplay.js file). That is why I didn't know what seeking is supposed to look like on butter while airplay is active.

All testing of airplay-protocol and bonjour has been verified by me using a separate application instead, which starts a basic HTTP server that AppleTV4 _can_ connect to. That's how I know the AirPlay commands implemented here work correctly, even though the AppleTV4 itself can't connect to _Butter's_ own HTTP server.

I've got the AppleTV4. I would need an AppleTV3 or older to be able to connect to Butter's streaming server, which I don't have.

I have no more time for this. The original airplay.js didn't even have pause and unpause, and didn't have seek at all. All commands are now implemented, plus more than it had in the past. Another contributor can pick up from this clean code and extend it with seeking in the future :) Perhaps after Butter's HTTP server is fixed!

Here's the HTTP server code I used for testing:
```javascript
  var server = http.createServer(function (req, res) {
    var stat = fs.statSync(filename)
    var size = stat.size
    var range = req.headers.range
    range = range && rangeParser(size, range)[0]

    res.writeHead(206, {
      'Access-Control-Allow-Origin': '*',
      'Accept-Ranges': 'bytes',
      'Content-Length': (range.end - range.start + 1),
      'Content-Type': contentType,
      'Content-Range': util.format('bytes %d-%d/%d', range.start, range.end, size)
    })

    fs.createReadStream(filename, range).pipe(res)
  })
```

AppleTV4 connects fine to that. It only fails to connect to Butter's own HTTP server.

Well, everything I can take care of is finished now. Let's merge and have fun and relax! ;-)

xaiki:
awesome, thanks, @watson does your code work on apple tv4 ?

xaiki:
```┌[~/src/Butter/butter-desktop:xaiki/kill-subtitles-in-generic]
└[furiloche] curl -vvvv http://127.0.0.1:49384/0 | head -1  
*   Trying 127.0.0.1...
* TCP_NODELAY set
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Connected to 127.0.0.1 (127.0.0.1) port 49384 (#0)
> GET /0 HTTP/1.1
> Host: 127.0.0.1:49384
> User-Agent: curl/7.51.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Accept-Ranges: bytes
< Content-Type: video/mp4
< transferMode.dlna.org: Streaming
< contentFeatures.dlna.org: DLNA.ORG_OP=01;DLNA.ORG_CI=0;DLNA.ORG_FLAGS=01700000000000000000000000000000
< Content-Length: 1153646658
< Date: Fri, 09 Dec 2016 01:11:38 GMT
< Connection: keep-alive
< 
{ [16384 bytes data]
ftypM4V M4V mp42isoi�moovlmvhd�OEw�OEw_�`@!iodsO�������trak\tkhd�OEw�OE�S�@��Amdia mdhd�OEw�OEw]�]�x�DhdlrvideMainconcept MP4 Video Media Handler��minfvmhd$dinfdref
              url Õstbl�stsd�avc1�HH
* Failed writing body (8192 != 16384)
* Curl_http_done: called premature == 1
* stopped the pause stream!
  0 1100M    0 81920    0     0  27.5M      0  0:00:39 --:--:--  0:00:39 39.0M
* Closing connection 0
curl: (23) Failed writing body (8192 != 16384)
┌[~/src/Butter/butter-desktop:xaiki/kill-subtitles-in-generic]
└[furiloche] curl -Ivvvv http://127.0.0.1:49384/0           
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 49384 (#0)
> HEAD /0 HTTP/1.1
> Host: 127.0.0.1:49384
> User-Agent: curl/7.51.0
> Accept: */*
> 
< HTTP/1.1 200 OK
HTTP/1.1 200 OK
< Accept-Ranges: bytes
Accept-Ranges: bytes
< Content-Type: video/mp4
Content-Type: video/mp4
< transferMode.dlna.org: Streaming
transferMode.dlna.org: Streaming
< contentFeatures.dlna.org: DLNA.ORG_OP=01;DLNA.ORG_CI=0;DLNA.ORG_FLAGS=01700000000000000000000000000000
contentFeatures.dlna.org: DLNA.ORG_OP=01;DLNA.ORG_CI=0;DLNA.ORG_FLAGS=01700000000000000000000000000000
< Content-Length: 1153646658
Content-Length: 1153646658
< Date: Fri, 09 Dec 2016 01:11:42 GMT
Date: Fri, 09 Dec 2016 01:11:42 GMT
< Connection: keep-alive
Connection: keep-alive

< 
* Curl_http_done: called premature == 0
* Connection #0 to host 127.0.0.1 left intact
``` that could be it

**Comment by xaiki** [comment]
so looking at the webtorrent code it has:
```js
    if (req.headers.origin) {
      res.setHeader('Access-Control-Allow-Origin', req.headers.origin)
    }
```
so that shoudl be fine, a tcpdump of negociation between your code and your TV could help us sort this out.

SteveJobzniak:
@xaiki Great to finally see this merged!

>awesome, thanks, @watson does your code work on apple tv4 ?

Yes, @watson's airplay-protocol and bonjour both work on the Apple TV4. They are sending the actual AirPlay and Bonjour commands correctly.

The problem is entirely in Butter's HTTP server. When I use the minimal HTTP server code I listed above, everything streams perfectly. But when I AirPlay via Butter the AppleTV screen flashes "black" as it starts playback but then goes back to the home screen when it fails to connect to Butter's server.

So yeah, the streaming bug itself is outside airplay.js.

I haven't got time for testing. Someone with AppleTV4 hardware and a Mac can easily do a tcpdump: https://support.apple.com/en-us/HT202013

I will move these relevant comments to your new issue.

xaiki:
we don't have an HTTP server, we use bittorrent.

SteveJobzniak:
@xaiki Ah you use webtorrent, right. But it sets up a streaming HTTP server where the device requests a URL to start streaming the file.

xaiki:
we give you back the webtorrent HTTP server:
```js
        createServer: function (port) {
            return new Promise(function (resolve) {
                var serverPort = parseInt((port || Settings.streamPort), 10);
                if (!serverPort) {
                    serverPort = this.generatePortNumber();
                }

                try {
                    this.torrentModel.get('torrent').createServer().listen(serverPort);

                    var url = 'http://127.0.0.1:' + serverPort + '/' + this.torrentModel.get('video_file').index;

                    this.streamInfo.set('src', url);
                    this.streamInfo.set('type', 'video/mp4');

                    resolve(url);
                } catch (e) {
                    setTimeout(function () {
                        return this.createServer(0).then(resolve);
                    }.bind(this), 100);
                }
            }.bind(this));
        },
```

SteveJobzniak:
@xaiki Ah, then the bug is in webtorrent. Because this is literally what happens:

`airplay.js -> play(streamingurl) -> sends AirPlay commant to AppleTV and gives the AppleTV the <streamingurl≥`.

There's nothing in airplay.js that affects playback. The AppleTV just receives a "please play this URL" command. And then it tries to connect to the URL it was given (which you say comes from webtorrent).

So the problem is in webtorrent.

Well, I have to return to real life now, this is more than what I have time for. Thanks for everything, take care and have a merry christmas! :)

drummwill:
has this been fixed? i am still unable to connect to my gen 4 aTV

PR #443, opened 2016-10-10
Cleaning remains

streger:
Removed 'PCT' from text line "Invalid PCT Database File Selected" to
make it more mainstream, and not special case.
...and so i did it ;)


xaiki:
thanks for you first contribution =) you should *only* touch `/src/app/lib/views/settings_container.js` on next app translation update it'll be exported to transifex and back in the languages.json files.

xaiki:
ideally we'd squash those commits out of the history, but we like you ;)

Issue #456, opened 2016-10-19
default subtitle error 100% of the time


vankasteelj:
Because of the opensubtitles-api update, setting a default language for sub will make the subtitle download fail.

The issue:
- api gives back an url
- we test that url to see if it ends in .zip or .srt
- it fails because it's an url, not a filename

The fix:
- figure it out

PR #460, opened 2016-10-22
fix for #456

vankasteelj:
- use a single function to download the targeted file, then detect if
  gz, zip or srt
- fix a case in hooks to avoid 'cannot read .thingy of null'
- fix opensubtitles-api on using .srt only


xaiki:
split promises work, split log -> console, but it looks good. just a lot to review with my poor little brain =P

PR #461, opened 2016-10-22
Choose default player on settings

vzamanillo:
- Hability to choose a default player ins settings view.

Hability to choose a default player ins settings view.

### DIFF
--- a/src/app/lib/views/settings_container.js
+++ b/src/app/lib/views/settings_container.js
@@ -61,6 +61,11 @@
                 }
             });
 
+            var el = $('#chosenPlayer');
+            App.Device.Collection.forEach(function (device) {
+                el.append(new Option(device.get('name'), device.get('id'), false, App.settings['chosenPlayer'] === device.get('id') ? true : false));
+            });            
+
             Mousetrap.bind('backspace', function (e) {
                 App.vent.trigger('settings:close');
             });
@@ -256,6 +261,9 @@
             case 'opensubtitlesUsername':
             case 'opensubtitlesPassword':
                 return;
+            case 'chosenPlayer':
+                value = $('option:selected', field).val();
+                break;
             default:
                 win.warn('Setting not defined: ' + field.attr('name'));
             }
--- a/src/app/templates/settings-container.tpl
+++ b/src/app/templates/settings-container.tpl
@@ -287,6 +287,13 @@
     <section id="playback">
         <div class="title"><%= i18n.__("Playback") %></div>
         <div class="content">
+            <span>
+                <div class="dropdown chosen-player">
+                  <p><%= i18n.__("Video Player") %></p>
+                  <select name="chosenPlayer" id="chosenPlayer" style="width: 160px;"></select>
+                  <div class="dropdown-arrow"></div>
+                </div>
+            </span>
             <span class="advanced">
                 <input class="settings-checkbox" name="alwaysFullscreen" id="alwaysFullscreen" type="checkbox" <%=(Settings.alwaysFullscreen? "checked='checked'":"")%>>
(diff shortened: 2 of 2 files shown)


vankasteelj:
Thanks, it works well, ready to merge. But shouldn't the dropdown disable external plays like airplay/dlna and such ? Because it will change on each reboot ?

vzamanillo:
Thank you, vankasteelj.

I can't test it because I have not external devices, anyway, I do not understand your question at all, In what conditions do you think the dropdown disable external plays? Is the device id for external players subject to change at each reboot?

xaiki:
i think it's ok to have external devices, we just need code to handle the case where they disappear.

xaiki:
(iirc we remember the last used player anyway)

vzamanillo:
If the choosen player does not exists anymore, the default selected device is set to 'local' so should not be a problem.

vankasteelj:
If this behavior is already set, then I don't have an issue. What I wanted to avoid, is for people to try to set "Chromecast" by default, then complain later : but it resetted and I have to manually set chromecast everytime, blah blah. Hence my suggestion to only display video player (in the settings, not in the "play now" dropdown) not remote devices

xaiki:
we could have a weird behaviour where we keep 2 players in config: one local one remote, this way if you set default to mpv then chromecast and chromecast isn't present, we fallback to mpv. not the clearest UX perhaps but i thought i'd share the idea =)

vzamanillo:
I don't like too much the idea of two configurations for default players, I think the default behaviour is better, one and only configuration (local or remote, does not matter) if the choosen player is not present set to default (local), the other way could be a little bit confusing for the user, but is my opinion.

xaiki:
yeah, what i'm ofering is to have an array in the conf:
```js
storePlayer: function (newplayer) {
savedPlayer = db.get('saved player')

if (newplayer.isLocal()) {
savedPlayer[0] = newplayer
} else {
savedPlayer[1] = newplayer
}

db.save(savedPlayer)
}
```
then when we select players we first try savedPlayer[1], then savedPlayer[0].

what do you think ?

vzamanillo:
OK, then I configure VLC, later configure Chromecast and I keep using Chromecast for six months and then Chromecast disappear because I disconnected it or whatever, open Butter got to movie details and I see... VLC as defautl player? what? why? I don't remember, and does not matter really, surely I will go for Chromecast to connect it again, the second player is irrelevant. Is not a bad idea, but I don't see the improvement.

vankasteelj:
IMHO you should only be able to select a local player as default (ie: in the settings). Don't display remote players there.

This way, for example:
- Default setting: 'butter' as default player. If I change that in the details view, it will only be remembered for the session, then fallback on 'butter' at app restart.
- Custom : 'VLC' as default player. If I change the player to 'butter' in the details view, then re-open the app, it will be set to VLC again. If I select Chromecast, it will use Chromecast until the next restart of the app. And if at app start, VLC isn't found, then default setting goes back to 'butter'.

vzamanillo:
But people will ask why their external players does not appear in the default player dropdown on settings but they can see in the dropdown of movie details and of course why they can not set their external device as default. I can't see a satisfying solution...  :confused:

streger:
I think that satisfying solution is to leave it as it was before. It behaved "smart", you choose before the player, and it remembers, and you don't have to deal with settings. Now, for me this complicates things and it raises more questions than simple solutions. And it clutters the UI. 
Having two things to solve one thing problem (that wasn't) is kind of too much.

xaiki:
actually, currently we store the last used player for the session.
i'm actually keen on having settings as small as possible (and i'm rewritting the DB code) so i guess in the end i'd like to drop this patch, and come back to this when i have the DB code merged then we can just remember the last used player.

vzamanillo:
Both of you are completly right.

PR #462, opened 2016-10-22
Default background on loading view

vzamanillo:
- If no background image or image could not be loaded when loading a movie or show we set the default background image on loading view.

If no background image or image could not be loaded we set the default background image on loading view.

### DIFF
--- a/src/app/lib/models/generic_collection.js
+++ b/src/app/lib/models/generic_collection.js
@@ -80,8 +80,14 @@
                                 }
                             }
 
-                            if (info.images.fanart) {
+                            if (info.images.full) {
                                 movie.backdrop = info.images.full;
+                            } else if (info.images.fanart && info.images.fanart.full) {
+                                movie.backdrop = info.images.fanart.full;
+                            } else if (movie.cover) {
+                                movie.backdrop = movie.cover;
+                            } else {
+                                movie.backdrop = 'images/bg-header.jpg';
                             }
                         } else {
                             win.warn('Unable to find %s (%s) on Trakt.tv', id, movie.title);
--- a/src/app/styl/views/loading.styl
+++ b/src/app/styl/views/loading.styl
@@ -20,6 +20,7 @@
         background-position-x 50%
         background-position-y 0px
         background-size cover
+        -webkit-filter: blur(15px) brightness(0.7)
 
     .loading-background-overlay
         width 100%
@@ -112,7 +113,7 @@
             background-color $ButtonBg
             -webkit-backface-visibility hidden
             transition all .5s
-            
+
             &:hover
                 background: $ButtonBgHover
                 text-decoration: none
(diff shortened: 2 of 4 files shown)


vankasteelj:
On a technical point, it has some flaws : the background changes abruptly (no more fadein in, because there already an image and the fadein was already applied to 'stock' image), and now, on every loading, there is either 1 image (if no real found, only stock), either 2 (stock, then real).

About the feel of this : I don't like it. I find this background image an abomination and prefer a color than some dirty jpeg coming from the early days of 0.3. But that's personnal

vzamanillo:
Thank you, you are right, the background image is near to an abomination :laughing:, technical flaws apart, my intention was to keep the same image background in the transition from the details to the load, with backdrop image or not.

xaiki:
another strategy is to use the poster as bg image if we don't find one. we do that in most providers and it works quite fine, especially with blur effects.

vzamanillo:
I think is a good idea, is better than the old background. I will take a look.

vankasteelj:
Blur ! Forgot about blur !

vzamanillo:
Blurred on movie or show details only or on loading view too?

xaiki:
it's already blured on details IIRC, on loading view i think it shouldn't be, but we should try and compare screenshots.

vzamanillo:
I think is not, or not enought, the solution should be easy, in the generic collection at [83](https://github.com/butterproject/butter-desktop/blob/master/src/app/lib/models/generic_collection.js#L83)

```javascript
if (info.images.fanart && info.images.full) {
	movie.backdrop = info.images.full;
}else{
	movie.backdrop = movie.cover;
}
```
and later apply the blur effect in the movie details styl

xaiki:
it's already blured on details IIRC, on loading view i think it shouldn't be, but we should try and compare screenshots.

xaiki:
i think it looks good enough. submit a PR and we'll tune it there ! thanks for the great work !

vzamanillo:
Thank you Xaiki.

xaiki:
i think it looks good enough. submit a PR and we'll tune it there ! thanks for the great work !

src/app/lib/models/generic_collection.js
```diff
@@ -80,8 +80,10 @@
                                 }
                             }
 
-                            if (info.images.fanart) {
+                            if (info.images.fanart && info.images.full) {
```
xaiki: what if fanart is defined and not images.full ?

xaiki: i think it should be:
```js
} if (images.full) {
 images.full
} else if (images.fanart) {
fanart
} else if cover {
cover
} else { 
handle error decide about something else.
}

File: src/app/styl/views/loading.styl [hunk 2]
```diff
@@ -20,6 +20,7 @@
         background-position-x 50%
         background-position-y 0px
         background-size cover
+        -webkit-filter: blur(15px) brightness(0.7)
```
vzamanillo: This is the only thing I'm not sure at all, without blur the transition is abruptly but on the other hand is cool to see without blur too...

PR #465, opened 2016-10-25
Multiple subtitle provider + dropdown availability

vzamanillo:
- If there are no subtitles available in any language we do not show the subtitles selector.

If there are no subtitles available in any language we do not show the subtitles selector.

### DIFF
--- a/src/app/templates/movie-detail.tpl
+++ b/src/app/templates/movie-detail.tpl
@@ -1,10 +1,10 @@
-<%  
-if(typeof backdrop === "undefined"){ backdrop = ""; }; 
-if(typeof synopsis === "undefined"){ synopsis = "Synopsis not available."; }; 
+<%
+if(typeof backdrop === "undefined"){ backdrop = ""; };
+if(typeof synopsis === "undefined"){ synopsis = "Synopsis not available."; };
 if(typeof runtime === "undefined"){ runtime = "N/A"; };
 if (genre) {
     for(var i = 0; i < genre.length; i++) {
-        genre[i] = i18n.__(genre[i]); 
+        genre[i] = i18n.__(genre[i]);
     }
 } else {
     var genre = [undefined];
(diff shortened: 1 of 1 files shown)


wallydz:
The issue is that Subtitle are not requested on Movie_details and it should be i was discussing that with @vankasteelj maybe we should call OpenSub on every Movie_details to have the Subtitle Dropdown useful again

vzamanillo:
I think should not be requested on movie details, maybe adding more providers and check for avaliability before.

xaiki:
we need subs providers =/ @amilajack did something like that in his experimental branch.
we need to load subs in the details view because of api rate limiting (and cache it agressively)

amilajack:
@xaiki 😂Its not a branch. Not even a fork. Worried about how we'll merge in changes. Probably gonna be a lot of conflicts.

vzamanillo:
But what is the real problem with the subtitles?

wallydz:
@vzamanillo Actually the Subtitle Dropdown on Movie_detail was used by Ysub we never adapted it to OpenSub , YIFI sub was allowing multiple Requests so we were able when loading movies we load Sub , now that we switch to Vodo for example we must use OpenSub and we cannot mass request like we were doing 

The idea would be the Request Opensub a list of available sub when user click on a Selected Movie

vzamanillo:
Ok, ok, it's clear now, I'll keep thinking about it, and how we could solve it.

xaiki:
@amilajack we need to separate your changes and better the butter-providers specs with them.

vzamanillo:
I was looking for alternative providers and found this

https://podnapisi.net/

Has a simple [HTTP API](https://www.podnapisi.net/forum/viewtopic.php?f=62&t=26164) and an [addon for Kodi ](https://github.com/amet/service.subtitles.podnapisi/tree/Gotham/service.subtitles.podnapisi)

I did not found nothing about rate limiting.

wallydz:
@vzamanillo would be interesting to have more Sub Provider , but from my old investigation Opensub was the most complete Lib

streger:
This is the MPC-HC SubtitlesProviders code, they use multiple sources:

https://github.com/mpc-hc/mpc-hc/blob/develop/src/mpc-hc/SubtitlesProvider.cpp

p.s. yes, the code is in c++ but you can take a look at least.

vankasteelj:
MPC-HC finds way less subs than we do...

streger:
Do you use nightly version or some stable. Because nightly version have auto-downlad and multi sources, it's in the code, just check it.

vankasteelj:
I have the version you can download on their website, so probably stable.

streger:
Try these, it's way better.
https://nightly.mpc-hc.org/

vankasteelj:
Anyway, the code you linked has:
- Opensubtitles -> done
- Yifysubtitles -> just have to port old code to a module or test https://github.com/oussama1598/yify-subs
- SubDB -> make a module out of http://thesubdb.com/api/ or test https://github.com/ka2er/node-subdb-api
- Titlovi -> http://en.titlovi.com/dev/ 
-  Pdonapisi -> https://www.podnapisi.net/subtitles/search/advanced there is no api at first sight, but xml scraping is recommended on their forums

vankasteelj:
I can do some of these pretty easily, subtitles is fun.

vankasteelj:
https://www.npmjs.com/package/yifysubtitles-api

vzamanillo:
Maybe sounds crazy, but why not decentralize the subs download (Creating and seeding a torrent from scratch, via magnet after download)? is this possible?

xaiki:
@vzamanillo: that's what we *should* do. everything should be a provider (a butter-provider) we need to port the metadata and subs to be modules (that you can configure). @amilajack has some work on that front but with another codebase.

xaiki:
didn't the sub server get merged ? can we close this ?

vankasteelj:
it has leads on multiple subs providers

xaiki:
there is only one commit left here, touching only the template, are you sure ?

vankasteelj:
I'm not rewriting the comments above in an issue.

PR #480, opened 2016-11-16
use external file for configuration

vankasteelj:
- benefits: easier to maintain & change to 'our' api keys
- inconvenients: a new js file to load + global variables

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -45,6 +45,7 @@
         "deleteFolder": true,
         "deleteCookies": true,
         "isVideo": true,
+        "api_defines": true,
         "minimizeToTray": true,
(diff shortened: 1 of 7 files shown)


xaiki:
we already have that, if you drop a butter-settings-x package in your node_modules we'll load that and overload settings with whatever you have there.

xaiki:
https://github.com/butterproject/butter-desktop/blob/master/src/app/bootstrap.js#L70

vankasteelj:
I'll have to check that, but what if it's loaded after the rest?

xaiki:
it is loaded last, so that it overrides all

Issue #313, opened 2016-03-09
Trakt todo-list (help welcomed)


vankasteelj:
- [x] introduce trakt.tv module
- [x] rewrite trakt provider (trakttv.js file) to use trakt.tv module instead (maintenance made easy)
- [ ] make trakttv.js a butter-provider
- [x] decide wheter localStorage is fitting (~~or use nedb like old implementation~~)
- [ ] display "episode" item instead of entire show
- [ ] fix the details (or remove the possibility of going to details at all)

https://github.com/butterproject/butter-desktop/commit/1417c20127db16c266772f526f1ad1eca7009dce

This commit rewrites entirely the watchlist and introduces the 'trakt.tv' package. 

There's a long list of things to do to have a proper trakt implementation. 

But first: @xaiki if everything needs to be rewritten, wouldn't it be a good time to split from core? Please assist on this matter.

YIFYtorrents:
who is behind trakt.tv. Don't get me wrong but can it be trusted?

vankasteelj:
It's a business runned by 3 guys I think.

xaiki:
https://github.com/butterproviders/butter-provider-trakt-watchlist
so part of the work is done here, it still lacks proper testing.

vankasteelj:
@xaiki it's only the watchlist though, not the entire trakt provider

xaiki:
sure, we'd need to split out from core, i'd say use localStorage.

PR #483, opened 2016-11-19
Rewrite trakt (using trakt.tv module)

vankasteelj:
Some fixes for https://github.com/butterproject/butter-desktop/issues/313

- issue was with watchlist syncing, fired too soon (provider was
  installed, but Settings wasn't loaded from DB yet)

### DIFF
--- a/src/app/lib/providers/trakttv.js
+++ b/src/app/lib/providers/trakttv.js
@@ -1,676 +1,241 @@
-(function (App) {
+(function(App) {
     'use strict';
 
-    var API_ENDPOINT = URI('https://api-v2launch.trakt.tv'),
-        CLIENT_ID = Settings.trakttv.client_id,
-        CLIENT_SECRET = Settings.trakttv.client_secret,
-        REDIRECT_URI = 'urn:ietf:wg:oauth:2.0:oob';
-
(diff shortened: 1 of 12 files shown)


src/app/lib/providers/trakttv.js
```diff
@@ -1,676 +1,241 @@
...
-                            season: item[i].episode.season,
-                            episode: item[i].episode.number,
-                        });
-                    }
-                    return calendar;
-                });
-        }
-    };
+        // auto sign-in when database is loaded
+        App.vent.on('db:ready', this.reAuthenticate.bind(this));
```
xaiki: is it any useful to have any trakt before db is ready ? shouldn't we make this a promise and just chain trakt (and everything that depends on that) on the loading of that promise ?

vankasteelj: To be honest, I don't know. Here Trakt is loaded pretty soon, the constructor is ready, and we simply log in when needed. Idk if by firing the entire provider once db is ready wouldnt make us loose some ms here and there.

xaiki: ok, let's merge this as is, we can always rewrite it with a promise later.

PR #484, opened 2016-11-20
uglify js/css files (saves 11Mb)

vankasteelj:
- gulp uglify task, for js/css files. 
- saves 11Mb of space (2.1Mb of app files, 8.9Mb of node_modules)
- not used by default, has to be called specifically
- incompatible with arrow functions, for some reason

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -443,15 +443,16 @@ gulp.task('uglify', () => {
                         let file = path.join(p,f);
                         switch (path.extname(f)) {
                             case '.js':
+                                var minified;
                                 try {
-                                    var minified = uglify.minify(file, {
+                                    minified = uglify.minify(file, {
                                         mangle: false
                                     });
                                 } catch (e) {
                                     looping = Date.now();
                                     return cb();
                                 }
-                                if (minified.code) {
+                                if (minified && minified.code) {
                                     fs.writeFile(file, minified.code, function (err) {
                                         if (!err) {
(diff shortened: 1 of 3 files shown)


xaiki:
i think we'd get better results (and smaller code) using webpack to do this...

vankasteelj:
it only saves space, it doesnt improve startup time (or it's unnoticable 1 or 2ms)

xaiki:
i've been thinking about this, i'm not sure we'd want it. because:
- we're an app that will need to DL about 1GB/h of data to be used, so 11Mb is less than a minute worth of DL.
- for less than a minute of update (that in the near future is not going to cost us server time as we're getting torrent updates) we're loosing the ability to trace user code when we encounter a bug.
- i think it really bloats our gulpfile...

anyway, let's pingpong some thoughts on this.

xaiki:
maybe this is a good starting point:
https://webpack.github.io/docs/usage-with-gulp.html
https://www.npmjs.com/package/gulp-webpack

SteveJobzniak:
My 2 cents: Getting nonsense function names in the console is terrible for debugging and user bug reports and is not worth saving 11mb.

SrPatinhas:
@SteveJobzniak well, for example my internet is 12mbps and regardless the speed, we should reduce/better solutions is always good practices. and maybe for the startup time do something different

vankasteelj:
Well, I'm putting an end to this little experiment.

PR #487, opened 2016-11-28
Xaiki/correct random

xaiki:
Random was broken, fix it.


vankasteelj:
commits b3d328a a421af8 need to go

vankasteelj:
see #489

PR #490, opened 2016-11-28
don't break the entire provider when one call breaks

vankasteelj:

### DIFF
--- a/src/app/lib/models/generic_collection.js
+++ b/src/app/lib/models/generic_collection.js
@@ -8,12 +8,12 @@
         var idsPromise = torrentsPromise.then(_.bind(torrentProvider.extractIds, torrentProvider));
         var promises = [
             torrentsPromise,
-            subtitle ? idsPromise.then(_.bind(subtitle.fetch, subtitle)) : true,
+            subtitle ? idsPromise.then(_.bind(subtitle.fetch, subtitle)).catch(function () { return false; }) : true,
             metadata ? idsPromise.then(function (ids) {
                 return Q.allSettled(_.map(ids, function (id) {
                     return metadata.getMetadata(id);
                 }));
-            }) : true
+            }).catch(function () { return false; }) : true
         ];
 
         //console.log('pre all', promises);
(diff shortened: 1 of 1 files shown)


src/app/lib/models/generic_collection.js
```diff
@@ -8,12 +8,12 @@
         var idsPromise = torrentsPromise.then(_.bind(torrentProvider.extractIds, torrentProvider));
         var promises = [
             torrentsPromise,
-            subtitle ? idsPromise.then(_.bind(subtitle.fetch, subtitle)) : true,
+            subtitle ? idsPromise.then(_.bind(subtitle.fetch, subtitle)).catch(function () { return false; }) : true,
```
xaiki: we need to at least log the error

src/app/lib/models/generic_collection.js
```diff
@@ -8,11 +8,19 @@
         var idsPromise = torrentsPromise.then(_.bind(torrentProvider.extractIds, torrentProvider));
         var promises = [
             torrentsPromise,
-            subtitle ? idsPromise.then(_.bind(subtitle.fetch, subtitle)) : true,
+
+            subtitle ? idsPromise.then(_.bind(subtitle.fetch, subtitle)).catch(function (err) {
+                win.warn('Cannot fetch subtitles (%s):', torrentProvider.name, err);
```
xaiki: let's not use win.warn as they don't give us good stack traces, please use console.error

src/app/lib/models/generic_collection.js
```diff
@@ -8,11 +8,19 @@
...
+                return false;
+            }) : true,
+
             metadata ? idsPromise.then(function (ids) {
                 return Q.allSettled(_.map(ids, function (id) {
                     return metadata.getMetadata(id);
                 }));
+            }).catch(function (err) {
+                win.warn('Cannot fetch metadata (%s):', torrentProvider.name, err);
```
xaiki: same

vankasteelj:
no we did not, you decided that yesterday and I won't start making direct console calls until the core is ported, it's a simple s//g and it's not a change I agree to make for a PR to a core code still using win.calls

PR #492, opened 2016-11-29
rewrite item.js

vankasteelj:
- more readable code in item.js (break in functions, rename variables, separate ui code & functionnal code)
- fixes the broken tooltips
- fixes addBookmark by calling the correct db for shows
- fixes 'mark as watched' for trakt
- clears the images mess (also freezes gif by using canvas)

### DIFF
--- a/src/app/lib/views/browser/item.js
+++ b/src/app/lib/views/browser/item.js
@@ -72,9 +72,10 @@
         },
 
         setModelStates: function () {
-            // Watched state
             var imdb = this.model.get('imdb_id');
             var itemtype = this.model.get('type');
+            
+            // Watched state
             var watched = false;
 
             if (itemtype.match('movie')) {
@@ -97,7 +98,7 @@
                 this.ui.bookmarkIcon.addClass('selected');
             }
 
-            if (itemtype.match('movie') && this.model.get('watched')) {
+            if (this.model.get('watched') && !itemtype.match('show')) {
                 this.ui.watchedIcon.addClass('selected');
 
                 switch (Settings.watchedCovers) {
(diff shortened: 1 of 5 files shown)


src/app/lib/views/browser/item.js
```diff
@@ -77,247 +71,216 @@
...
+            var imdb = this.model.get('imdb_id');
+            var itemtype = this.model.get('type');
+            var provider = App.Providers.get(this.model.get('provider'));
+
+            return provider.detail(imdb, this.model.attributes).then(function (data) {
+                data.provider = provider.name;
+
+                if (itemtype === 'show') {
+                    promisifyDb(db.shows.find({
+                        imdb_id: imdb.toString(),
```
xaiki: i'd rather have no direct calls to db here, let's move that into db and call a helper function from here ?

src/app/lib/views/browser/item.js
```diff
@@ -77,247 +71,216 @@
...
+        removeBookmarked: function () {
+            var imdb = this.model.get('imdb_id');
+            var itemtype = this.model.get('type');
+            var dbCall = (function () {
+                if (itemtype.match('show')) {
+                    return 'deleteTVShow';
+                } else {
+                    return 'deleteMovie';
+                }
+            }());
```
xaiki: earlier you used an implicit if statement for the same thing, why use a closure here ?

vankasteelj: because it was movies only. Here we can add 'short' or w/e future thing we implement.

Above : watched: Boolean
Here: itemtype: String

PR #493, opened 2016-11-29
switch win{debug,info,warn,error} to console calls

vankasteelj:


src/app/lib/models/generic_collection.js
```diff
@@ -62,7 +58,7 @@
                         movie.subtitle = subtitles[id];
                     }
 
-                    if (metadatas) {
+                    if (metadatas && id) {
```
xaiki: haha that got in =P

src/app/lib/providers/generic.js
```diff
@@ -44,29 +44,28 @@
...
         if (cache[name]) {
-            win.info('Returning cached provider', name);
             return cache[name];
         }
 
         var provider = getProviderFromRegistry(config.name);
         if (!provider) {
             if (installProvider(require('butter-provider-' + config.name))) {
-                win.warn('I loaded', config.name, 'from npm but you didn\'t add it to your package.json');
+                console.warn('I loaded', config.name, 'from npm but you didn\'t add it to your package.json');
```
xaiki: note to self: this is where we handle npm providers

src/app/lib/views/main_window.js
```diff
@@ -138,14 +138,6 @@
...
         },
 
-        showSubtitles: function (model) {
-            win.debug('Show subtitles', model);
-            var s = new App.View.Subtitles({
-                model: model
-            });
-            s.render();
-        },
-
```
xaiki: is that right ?!

xaiki:
another great PR !

src/app/lib/providers/generic.js
```diff
@@ -17,14 +17,14 @@
...
+            return console.error(PO, PO.prototype.config, 'doesnt have a name');
         }
 
         if (registry[name]) {
-            return win.error('double definition of', name, PO, PO.prototype.config, 'is the same as', registry[name]);
+            return console.error('double definition of', name, PO, PO.prototype.config, 'is the same as', registry[name]);
         }
 
-        win.debug('added', name, 'to provider registry');
+        //console.log('Added %s to provider registry', name);
```
vankasteelj: It's spam to me. If you disagree, uncomment it: what bothers me is that it's repeated every load, without any real need for it. I mean, it's in package.json, it should be loaded, I don't see why the log is needed (except when working on it). Logging isn't commenting

src/app/lib/views/player/player.js
```diff
@@ -343,8 +343,6 @@
 
         onShow: function () {
             $('#header').removeClass('header-shadow').hide();
-            // Test to make sure we have title
-            win.info('Watching:', this.model.get('title'));
```
vankasteelj: I don't see why exactly, we see the title on the app, the log doesnt add anything

src/app/lib/providers/generic.js
xaiki: it helps me debug issues when loading providers... please keep it.

src/app/lib/views/player/player.js
xaiki: that one i don't really care about, but the point of logging is that we put it there for some reason at some point.

xaiki:
pushed the provider logs changes to master

PR #498, opened 2016-12-01
Vkj/dbclick player

vankasteelj:
- on click, wait to see if a doubleclick event passes
- if a dblclick is detected, ignore initial click // else proceed with click
- allows to not spam pause/play on doubleclick
- move ui.{play,pause} to their own function

Note: using `$('#video_player').dblclick` over `this.player.tech.dblclick` because we can then bypass the mouse being hidden (and tech is hidden too) when fullscreen

### DIFF
--- a/src/app/lib/views/player/player.js
+++ b/src/app/lib/views/player/player.js
@@ -414,7 +414,7 @@
             this.player = this.video.player();
 
             // Better handle click/dblclick events for play/pause/fs
-            this.player.click = {count: 0, dblclick: false}; // init conf
+            this.player.click = 0;
             this.player.tech.off('mousedown'); // stop listening to default ev
             this.player.tech.on('mouseup', this.onClick.bind(this));
             $('#video_player').dblclick(this.onDbClick.bind(this));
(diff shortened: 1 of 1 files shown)


src/app/lib/views/player/player.js
```diff
@@ -482,6 +469,33 @@
...
+            this.player.click.count++;
+
+            if (this.player.click.count === 1) { // it's the first time we click
+                setTimeout(function () { // wait for double click
+                    if (!this.player.click.dblclick) { // if we didn't catch dbclick
+                        $('.vjs-play-control').click();
+                    }
+                    this.player.click.count = 0;
+                    this.player.click.dblclick = false;
+                }.bind(this), 300);
```
xaiki: make that a configurable variable either in settings or at least on top of the file

src/app/lib/views/player/player.js
```diff
@@ -482,6 +469,33 @@
...
+            } else { // we already clicked, reset
+                this.player.click.count = 0;
+                this.player.click.dblclick = false;
+            }
+        },
```
xaiki: this will work, but it's not very elegant, as i suggested on slack we may want to have a common handler for 'click' and 'double click' events that is `_.throttle`d to get the edge one. what do you think ?

src/app/lib/views/player/player.js
```diff
@@ -482,6 +469,33 @@
             });
         },
 
+        onClick: function (e) {
+            if (!this.player.click) {
```
xaiki: that can go after the init of player so we don't check it on every onClick

src/app/lib/views/player/player.js
```diff
@@ -482,6 +469,33 @@
...
+                this.player.click = {};
+                this.player.click.count = 0;
+            }
```
xaiki: this can be:
```js
this.player.click = {
count: 0,
dblclick: false
}```

PR #508, opened 2016-12-03
backport from forks

vankasteelj:
pick useful updates from forks

### DIFF
--- a/src/app/lib/device/chromecast.js
+++ b/src/app/lib/device/chromecast.js
@@ -55,18 +55,20 @@
                     title: streamModel.get('title').substring(0,50)
                 };
             }
-            console.log('Chromecast: play ' + url + ' on \'' + this.get('name') + '\'');
-            console.log('Chromecast: connecting to ' + this.device.host);
+
+            console.log('Chromecast: play %s on %s', url, this.get('name'));
+            console.log('Chromecast: connecting to %s', this.device.host);
 
             self.device.play(url, media, function (err, status) {
                 if (err) {
-                    console.error('chromecast.play error: ', err);
+                    console.error('Chromecast.play() error: ', err);
                 } else {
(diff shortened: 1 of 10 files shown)


src/app/lib/device/chromecast.js
```diff
@@ -88,36 +90,17 @@
...
         },
 
         forward: function () {
-            this.seek(30);
+            this._seek(this.get('loadedMedia').currentTime + 30);
         },
 
         backward: function () {
-            this.seek(-30);
+            this._seek(this.get('loadedMedia').currentTime - 30);
```
xaiki: implement that in _seek ? or seek ?

src/app/lib/views/movie_detail.js
```diff
@@ -7,7 +7,8 @@
 
         ui: {
             bookmarkIcon: '.favourites-toggle',
-            watchedIcon: '.watched-toggle'
+            watchedIcon: '.watched-toggle',
+            selected_lang: '.selected-lang'
```
xaiki: no no no no

xaiki: this belongs into the dropdown code.

src/app/lib/device/chromecast.js
vankasteelj: no, read the 3 functions using _seek, it's not feasible

src/app/lib/views/movie_detail.js
vankasteelj: I'm backporting from forks, it does make things better than without it (hence why they added it)

src/app/lib/device/chromecast.js
xaiki: then do seek, if you don't want to implement this i'll do it but don't complain i touch your branches...

PR #514, opened 2016-12-04
Xaiki/provider opts

xaiki:
provider ops rebased, it seems to work.
makes lang_dropdown do what it should.
i'm seeing very slow metadata times, but it shouldn't be related to what this touches...


src/app/global.js
```diff
@@ -5,6 +5,9 @@ var
     inherits = require('util').inherits,
     Q = require('q'),
 
+    // es6 pollyfills
+    objectAssign = require('es6-object-assign').assign,
```
vankasteelj: Since we moved to nw18, we don't need a polyfill I think.

src/app/lib/models/generic_collection.js
```diff
@@ -52,7 +52,8 @@
 
                         return;
                     }
-                    movie.provider = torrentProvider.name;
+
+                    movie.provider = torrentProvider;
```
vankasteelj: Should remain torrentProvider.name, torrentProvider is an object. Or if you want to change that (see comment below), there is more than just here

src/app/lib/views/lang_dropdown.js
```diff
@@ -1,47 +1,80 @@
 (function (App){
     'use strict';
 
+    var objectAssign = require('es6-object-assign').assign;
```
vankasteelj: There is no need to defined twice objectAssign (if defined in global, it's already accessible)

src/app/lib/views/browser/item.js
```diff
@@ -164,7 +158,7 @@
             var realtype = this.model.get('type');
             var itemtype = realtype.replace('bookmarked', '');
             var modelType = itemtype.charAt(0).toUpperCase() + itemtype.slice(1); // 'Movie', 'Show'
-            var provider = App.Providers.get(this.model.get('provider'));
+            var provider = this.model.get('provider');
```
vankasteelj: Related to the above comment: if the backbone model has to support the entire provider functions (before there was just a string attached), isnt that memory-consuming? I mean, the provider is quite an heavy object to pass attached to a model, isnt it? Anyway, if this is really wanted, then I think there are other places where model.attributes.provider is still using a string (all should be ported, for consistency)

src/app/templates/play-control.tpl
```diff
@@ -0,0 +1,36 @@
+<div class="play-control">
```
vankasteelj: ok, this is split in "component", great, but there is no css attached to it. Shouldnt there be a play-control.styl?

vankasteelj:
Remarks (while using)

1) Visually: at 960x580 (min WxH for the window), it's ugly: 
nw_2016-12-04_11-33-50

2) backdrop is broken on moviedetails (stays black)

3) provider.detail is not a function error, probably linked to your changes to provider above.
- enter details
- close details
- you can't enter a 2nd time

src/app/lib/views/play_control.js
```diff
@@ -0,0 +1,283 @@
...
+        switchAudio: function (lang) {
+            var audios = this.model.get('audios');
+
+            if (audios === undefined || audios[lang] === undefined) {
+                lang = 'none';
+            }
+
+            this.audio_selected = lang;
+
+            console.info('Audios: ' + this.audio_selected);
```
vankasteelj: this is called twice on every load, coulnt figure out why

src/app/global.js
```diff
@@ -5,9 +5,6 @@ var
     inherits = require('util').inherits,
     Q = require('q'),
 
-    // es6 pollyfills
-    objectAssign = require('es6-object-assign').assign,
```
xaiki: cool, did we include that in package.json?

vankasteelj: There is no need, Object.assign is there in nwjs >= 13, through updated node

PR #518, opened 2016-12-05
Xaiki/kill subtitles in generic

xaiki:
implement subs fetching in item.js, it looks like we should move it to the view details class, but it sits there nicely anyway.

not sure we shouldn't remove this: 
```js
            // bookmarked movies are cached
            if (realtype === 'bookmarkedmovie') {
                return App.vent.trigger('movie:showDetail', this.model);
            }
```
as we'd miss subtitles for bookmarked movies but it's probably part of another bigger smarter PR.

### DIFF
--- a/src/app/lib/providers/opensubtitles.js
+++ b/src/app/lib/providers/opensubtitles.js
@@ -29,11 +29,15 @@
         for (var lang in data) {
             data[lang] = data[lang].url;
         }
-        return Common.sanitize(data);
+        return {subtitle: Common.sanitize(data)};
     };
 
-    OpenSubtitles.prototype.fetch = function (queryParams) {
-        queryParams.extensions = ['srt'];
+    OpenSubtitles.prototype.detail = function (id, attrs) {
+        var queryParams = {
+            imdbid: id,
+            extensions: ['srt']
+        };
+
         return openSRT.search(queryParams)
             .then(formatForButter);
     };
--- a/src/app/lib/providers/opensubtitles.js
+++ b/src/app/lib/providers/opensubtitles.js
@@ -29,17 +29,24 @@
         for (var lang in data) {
             data[lang] = data[lang].url;
         }
-        return {subtitle: Common.sanitize(data)};
+        return Common.sanitize(data);
     };
 
-    OpenSubtitles.prototype.detail = function (id, attrs) {
-        var queryParams = {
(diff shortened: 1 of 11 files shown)


src/app/lib/views/browser/item.js
```diff
@@ -1,6 +1,8 @@
 (function (App) {
     'use strict';
 
+    var Q = require('q');
```
vankasteelj: this is already defined in global.js

src/app/lib/providers/opensubtitles.js
```diff
@@ -29,11 +29,15 @@
         for (var lang in data) {
             data[lang] = data[lang].url;
         }
-        return Common.sanitize(data);
+        return {subtitle: Common.sanitize(data)};
     };
 
-    OpenSubtitles.prototype.fetch = function (queryParams) {
-        queryParams.extensions = ['srt'];
+    OpenSubtitles.prototype.detail = function (id, attrs) {
```
vankasteelj: No. That's a bad idea. We don't search by imdb only, we use filename on dropped files and so on. Let the module handle what queries it can accept and what it can't. It was written correctly before.

vankasteelj: Or you need to write two functions, one .detail wrapping the detail call, one .fetch to actually fetch

src/app/lib/views/lang_dropdown.js
```diff
@@ -22,9 +21,7 @@
             if (this.hasNull) {
                 this.values = Object.assign({}, {none: undefined}, this.values);
                 this.model.set('values', this.values);
-            }
-
-            if (!this.selected && this.values) {
+            } else if (!this.selected && this.values) {
```
vankasteelj: are you sure? if hasNull, then the event wont be triggered

src/app/lib/providers/opensubtitles.js
xaiki: oh, the former implementation just took queryparams (that is id) and passed it to the module, it had a bug trying to extend it, OS is your code i'm not sure what's the best way to handle this. want to take it on from here ?

xaiki: thanks ! i think we need to remove some OS/sub fetching from streamer too btw

PR #519, opened 2016-12-05
Xaiki/css flags

xaiki:


bower.json
```diff
@@ -1,27 +1,54 @@
...
-    }
+  "name": "butter",
+  "dependencies": {
+    "marionette": "^2.4.1",
+    "bootstrap": "~3.3.7",
+    "video.js": "4.11.4",
+    "videojs-youtube": "1.2.10",
+    "mousetrap": "~1.6.0",
+    "font-awesome": "^4.6.3",
+    "flag-icon-css": "^2.7.0"
```
vankasteelj: why is that in both bower & npm?

xaiki: that's a mistake, clearly, i'd like to have it in npm, but i'm not sure how to include it...

xaiki:
i haven't removed the old ones because some i still use, and we need to go through all the languages to find the ones that we'd want to keep (chinese has a split china/tw flag for instance).

this PR started because now that we get subs, a lot of flags were empty.

bower.json
vankasteelj: I really don't know^^ worth some searching though.

src/app/language.js
```diff
@@ -158,7 +176,8 @@ App.Localization.langcodes = {
     },
     'ca': {
         name: 'Catalan',
-        nativeName: 'Català'
+        nativeName: 'Català',
+        flag: 'catalan'
```
vankasteelj: why not flag:'ca' or 'es-ca'? it's a little weird to have some flags using a 2 letter langcode, other full name. Not really important, but might worth considering

xaiki: so 2 letter codes come from the flag library, the full names are the ones we provide. for arabic it was confusing as ar = argentina ? es-ar ? ar... some would say that catalan has nothing to do with spanish... that said we have catalan and occitan, and i'm pretty sure fr-oc is wrong.

wallydz: AR is the ISO code for arabic lang
.ar is the TLD for argentina :p

vankasteelj:
it's ugly as hell

xaiki:
closing this, on twitter people asked for flags, we'll wait to see what @abad says with design.

PR #527, opened 2016-12-08
event handling

### LINKED ISSUE #520: Events arent called off
We need to check all events and unload them (App.vent.off) if they're reloaded on user interaction, to avoid having the same event fired multiple times (I noticed some being fired up to 7 times at once)

vankasteelj:
- call off events that need it (closes #520)
- rewrite on('app:started') event to avoid dupes
- enhance -m option: avoid screen flickering when the app opens minimized
  due to delay between win.show & win.hide

### DIFF
--- a/src/app/lib/streamer.js
+++ b/src/app/lib/streamer.js
@@ -52,6 +52,15 @@
             }.bind(this)).then(this.waitForBuffer.bind(this)).catch(this.handleErrors.bind(this));
         },
 
+        restart: function () {
+            var torrent = this.webtorrent.torrents[0].magnetURI;
+            this.webtorrent.remove(torrent, function (err) {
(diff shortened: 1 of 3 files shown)


src/app/lib/streamer.js
```diff
@@ -509,9 +518,9 @@
         },
     };
 
-    var streamer = new WebTorrentStreamer();
+    App.Streamer = new WebTorrentStreamer();
```
vankasteelj: seems I forgot about that. It's just to have access to the streamer from console, doesn't change anything

xaiki:
a lot of changes that really should be split into many commits...

xaiki:
i've split your commits into smaller ones in xaiki/520

PR #531, opened 2016-12-09
Greek translation improvements

sopex:
Fixes # .

- The translation now makes sense
- 
- 

I improved the greek translation. Some translations didn't make sense.


xaiki:
hey thanks for your PR ! but translations are handled throuh transifex ! please coordinate with @insertc0in for those

sopex:
@xaiki Do you prefer to make the some extra changes to transifex or here?

xaiki:
sopex, i prefer we do them on transifex as it overwrittes our translations. 
that said i need to figure out how that works to be honest

sopex:
@xaiki Just to know every single word of greek translation is in the wrong form so I have to change the 80% of the file and its a bit difficult in transifex. So if you can deactivate greek from transifex to don't overwrite the files.

insertc0in:
@sopex Thanks for your contribution. 
Unfortunately, as @xaiki has explained, translations are not directly commited to the source. Instead we ask translators/contributors to use Transifex, which is from where we pull them. Looking at your commit I see that you made 13 changes, which does not seem that big of a stretch to do in Transifex. We would appreciate a ton if you could do it there, we don't really want to disturb the current feng shui we have going on. Thanks for your understanding.

xaiki:
@insertc0in can you push this to transifex and make sure that PR is closed when done please ?

insertc0in:
@xaiki, currently I am still waiting for @sopex to reply to my request about implementing it in Transifex. I don't see a reason to push 13 strings into Transifex, worst case I'll copy/paste them myself since it would bring less complications.

xaiki:
great thanks, let's say that if we don't get an update by next week we'll handle that this way ?

insertc0in:
Sure, no problem.

insertc0in:
I've copied the changes in this PR into Transifex.
Further changes MUST be handled through there. Thanks @sopex for the contribution.

PR #536, opened 2016-12-12
Fix External Device Source

### LINKED ISSUE #528: streaming is broken on Apple TV4
see @SteveJobzniak's comments on #437 
we'd need some tcpdumps to see what's happening as nobody owns that hardware.

team-pct:
Fixes #528 

- Disable Comparaison and picking up all interfaces
- Using Network-Adresss module to directly set up IPv4 Source

Disable Comparaison and picking up all interfaces
Using Network-Adresss module to directly set up IPv4 Source
Related Issue #528

### DIFF
--- a/src/app/lib/device/generic.js
+++ b/src/app/lib/device/generic.js
@@ -3,27 +3,6 @@
 
     var self;
 
-    // Supports both IPv4 and IPv6 comparison
-    var _sequentialPartsInCommon = function (ip1, ip2) {
-        var separator = (ip1.indexOf('.') > -1) ? '.' : ':';
-        var ip2Parts = ip2.split(separator),
-            partsCount = 0;
-        ip1.split(separator).every(function (ip1Part, idx) {
-            var isEqual = (ip1Part === ip2Parts[idx]);
-            if (isEqual) {
-                ++partsCount;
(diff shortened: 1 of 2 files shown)


src/app/lib/device/generic.js
```diff
@@ -91,26 +70,8 @@
...
-                            ips.push(details.address);
-                        }
-                    });
-                }
-                var deviceIp = this.selected.get('address');
-                console.log('Device IP: ' + deviceIp);
-                console.log('Available IPs: ' + JSON.stringify(ips));
-                var srcIp = _getClosestIP(ips, deviceIp);
-                console.log('%s picked for external playback', srcIp);
+                var srcIp = require('network-address')();
```
xaiki: put this at the top of the file:
```js
var NetworkAddress = require('network-address');
```
and here you:
```js
var srcIp = NetworkAddress()
```

SteveJobzniak:
@team-pct This would break if someone has a future network with only IPv6 (it could happen? 💃 ).

Better to focus energy on this instead, which makes it possible to set the IP family (v4/v6) per device type:
https://github.com/butterproject/butter-desktop/pull/537

PR #537, opened 2016-12-12
make ipFamily configurable in device

### LINKED ISSUE #528: streaming is broken on Apple TV4

xaiki:
Fixes #528 .

please test !!!

### DIFF
--- a/src/app/lib/device/airplay.js
+++ b/src/app/lib/device/airplay.js
@@ -21,7 +21,8 @@
     var AirPlay = App.Device.Generic.extend({
         defaults: {
             type: 'airplay',
-            typeFamily: 'external'
+            typeFamily: 'external',
+            ipFamily: 'IPv4'
         },
 
         _makeID: makeID,
--- a/src/app/lib/device/airplay.js
+++ b/src/app/lib/device/airplay.js
@@ -22,7 +22,7 @@
         defaults: {
             type: 'airplay',
             typeFamily: 'external',
-            ipFamily: 'IPv4'
+            ipFamily: App.Device.IP_FAMILY.IPV4
         },
 
         _makeID: makeID,
--- a/src/app/lib/device/generic.js
+++ b/src/app/lib/device/generic.js
@@ -2,6 +2,8 @@
     'use strict';
 
     var self;
+    var dns = require('dns'),
+        os = require('os');
 
     // Supports both IPv4 and IPv6 comparison
     var _sequentialPartsInCommon = function (ip1, ip2) {
@@ -45,27 +47,42 @@
              * Supports IPv4 and IPv6.
              */
             if (this.get('typeFamily') !== 'external') {
-                return streamModel;
+                return Promise.resolve(streamModel);
             }
 
             //console.warn('External Device ', this.selected);
             var ips = [],
-                ifaces = os.networkInterfaces();
+                ifaces = os.networkInterfaces(),
+                ipFamily = this.get('ipFamily');
             for (var dev in ifaces) {
                 ifaces[dev].forEach(details => {
-                    if (!details.internal && details.family.match(this.ipFamily)) {
+                    if (!details.internal && details.family.match(ipFamily)) {
                         ips.push(details.address);
                     }
                 });
             }
-            var deviceIp = this.selected.get('address');
-            console.log('Device IP: ' + deviceIp);
-            console.log('Available IPs: ' + JSON.stringify(ips));
-            var srcIp = _getClosestIP(ips, deviceIp);
-            console.log('%s picked for external playback', srcIp);
-            var smIp = streamModel.get('src');
-            streamModel.set('src', smIp.replace('127.0.0.1', srcIp));
-            return streamModel;
+
+            return new Promise((resolve, reject) => {
+                // XXX(xaiki): we'd need a better way to check what we want
+                var options = ipFamily.mach('4') ? {family: 4} :
+                    ipFamily.mach('6') ? {family: 6} :
+                    {};
+
+                dns.lookup(this.selected.get('address'), options,
+                           (err, deviceIp, family) => {
+                    if (err) {
+                        return reject(err);
+                    }
+
+                    console.log('Device IP: ' + deviceIp);
+                    console.log('Available IPs: ' + JSON.stringify(ips));
+                    var srcIp = _getClosestIP(ips, deviceIp);
+                    console.log('%s picked for external playback', srcIp);
+                    var smIp = streamModel.get('src');
+                    streamModel.set('src', smIp.replace('127.0.0.1', srcIp));
+                    return resolve(streamModel);
+                });
+            });
         }
     });
 
@@ -121,8 +138,8 @@
                 this.selected = this.models[0];
             }
 
-            streamModel = this.selected.setIP(streamModel);
-            return this.selected.play(streamModel);
+            this.selected.setIP(streamModel)
+                .then(this.selected.play.bind(this));
         },
 
         setDevice: function (deviceID) {
--- a/src/app/lib/device/generic.js
+++ b/src/app/lib/device/generic.js
@@ -29,6 +29,7 @@
             id: 'local',
             type: 'local',
             typeFamily: 'internal',
+            ipFamily: '.*', // a regex to match the ip.family
             name: Settings.projectName
         },
         play: function (streamModel) {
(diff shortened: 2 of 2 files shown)


vankasteelj:
this breaks dlna.

xaiki:
any more info ?

vankasteelj:
cannot read mach of undefined

xaiki:
hum, you're on windows ? dlna works here, can you give me an output of os.networkInterfaces() ?

vankasteelj:
Error on play

``` 
`Uncaught (in promise) TypeError: Cannot read property 'mach' of undefined
    at Promise (chrome-extension://afkpagbcdgaomobnegnmhijblpoeanna/src/app/lib/device/generic.js:67:39)
    at n.setIP (chrome-extension://afkpagbcdgaomobnegnmhijblpoeanna/src/app/lib/device/generic.js:65:20)
    at n.startDevice (chrome-extension://afkpagbcdgaomobnegnmhijblpoeanna/src/app/lib/device/generic.js:141:27)
    at n.streamReady (chrome-extension://afkpagbcdgaomobnegnmhijblpoeanna/src/app/lib/views/main_window.js:454:35)
    at _ (chrome-extension://afkpagbcdgaomobnegnmhijblpoeanna/src/app/vendor/backbone/backbone-min.js:1:3942)
    at m (chrome-extension://afkpagbcdgaomobnegnmhijblpoeanna/src/app/vendor/backbone/backbone-min.js:1:3728)
    at f (chrome-extension://afkpagbcdgaomobnegnmhijblpoeanna/src/app/vendor/backbone/backbone-min.js:1:1563)
    at c.l.trigger (chrome-extension://afkpagbcdgaomobnegnmhijblpoeanna/src/app/vendor/backbone/backbone-min.js:1:3608)
    at WebTorrentStreamer.watchState (chrome-extension://afkpagbcdgaomobnegnmhijblpoeanna/src/app/lib/streamer.js:377:26)
    at chrome-extension://afkpagbcdgaomobnegnmhijblpoeanna/src/app/vendor/underscore/underscore-min.js:5:8229
```
network:
nw_2016-12-12_23-32-31

SteveJobzniak:
This looks great! Thanks for doing this! I suggested some fixes which should make it work. 🍰

xaiki:
@vankasteelj i pushed a fix for your issue, for some reasons defaults are not interhitted..

vankasteelj:
I still encounter the same issue

vankasteelj:
Ok, so, there are many, many issues:

1) Object.assign doesnt work in this case, it messes the object (maybe because it's a backbone model, not an object?)
2) this.selected is undefined
3) ipFamily isn't set anywhere except in airplay
4) .mach should be .match

SteveJobzniak:
@vankasteelj +1 everything that you said. And don't miss my code review comments @xaiki ;)

xaiki:
ok, i see, i went for a simpler solution, reworked the patchset, should all work now.

SteveJobzniak:
@xaiki Okay, those changes look good! 👍 

Go up above to "SteveJobzniak started a review" and please implement the two comments (because the code you added isn't obvious on 1st glance, so those sections need comments), and if you agree with my proposed fix for `// XXX(xaiki): we'd need a better way to check what we want` then add that too (I posted some replacement code for it in my review). 🍰

xaiki:
@SteveJobzniak: i can't see your review for some reason ? did you submit it ?

SteveJobzniak:
@xaiki Oops, I didn't hit Submit. 😨 😆

xaiki:
thanks @SteveJobzniak integrated your comments ! the other part doesn't make anything much better.

SteveJobzniak:
@xaiki Good to see! :)

About the other change: What doesn't make things better? Your code is bugged more than you might realize. ;)

Currently you are using:

```javascript
// XXX(xaiki): we'd need a better way to check what we want
var options = ipFamily.match('4') ? {family: 4} : ipFamily.match('6') ? {family: 6} : {}
```

Bugs:

1. ipFamily is a regex. Don't look for plain numbers in it. That's no problem today but maybe later.
2. The code risks breaking IPv6 networks.

I will explain:

Imagine that a device library is set up like this:

```javascript
dlna.js:
ipFamily = '(4|6)'
```

Then your code runs:

```javascript
var options = ipFamily.match('4') ? {family: 4} : ipFamily.match('6') ? {family: 6} : {}

result:
options = {4}
```

Options becomes ipv4-only, because the device pattern was marked as supporting both types of networks, and it checked for the number "4" first and decided to look for ONLY IPv4 networks when scanning for the host of this Butter device-type.

Now imagine that a user of Butter has only IPv6 on their network. And you are telling Node to do a IPv4 DNS query.

Result: Failed to look up IPv4 IP on this network. 🎉 

--------

My solution was to actually log the true Node.js names of all families that matched:

```javascript
              var ips = [],
                  ifaces = os.networkInterfaces(),
                  ipFamily = this.get('ipFamily'),
 +                matchedFamilies = {};
              for (var dev in ifaces) {
                  ifaces[dev].forEach(details => {
                      if (!details.internal && details.family.match(ipFamily)) {
                          ips.push(details.address);
 +                        matchedFamilies[details.family] = true; // string, either "IPv4" or "IPv6"
                      }
                  });
              }
```

Then we can truly know which families were enabled on YOUR local network (NOT just what the DEVICE.js lib says that it wants). Then we tell Node to do a IPv4+IPv6 lookup, or IPv4, or IPv6, all based on what the user's network supported:

```javascript
// lookup the device's IPv4, IPv6 or both.
// fallback to looking for IPv4 if no valid families found.
var options = ( (matchedFamilies["IPv4"] && matchedFamilies["IPv6"]) ? {} :
  matchedFamilies["IPv6"] ? {family: 6} : {family: 4} );
```

I think this code should be merged too, unless you have an even better idea? :)

xaiki:
so just to explain, i'd rather make it an enum, so that you especifically ask for IPV4 IPV6 or ALL, that enum shoud be in the node lib btw, it's not. the regex is really an iplementation detail and i don't want devices to depend on that.

vankasteelj:
It works for me now, even though the "stop" button on DLNA says "filetype not accepted" instead of just closing the stream, but that's a separate issue

SteveJobzniak:
@xaiki That enum looks _fantastic_ and fixes the potential bug if a device had used a "(4|6)" regex. Looks great. 👍 

Excellent work on this patch. Thank you for helping solve this AppleTV issue (and any other IPv4-only device we ever add). ✌️

SteveJobzniak:
@xaiki @vankasteelj By the way, I just looked at the whole merge and saw a small problem in the final merge.

https://github.com/butterproject/butter-desktop/commit/8b2d3ff7d088090a490f99debd038e64e6419aa6#diff-2bad2338b077ec7a37a5d8726509ccc3R40

That line for generic devices should be changed from:

`ipFamily: '.*', // a regex to match the ip.family`

To:

`ipFamily: IP_FAMILY_ENUM.ALL,`

Right?

SteveJobzniak:
@xaiki @vankasteelj Alright, I have tested this commit now.

## Background info about the network:

**MacBookPro**
10.0.1.248

**AppleTV**
10.0.1.164

## Testing:

I added some debug output which prints the URL that is being fed to airplay.js play() function, meaning the URL we will be sending to the AppleTV and saying "here, please play this URL!". So what you see in `THEURL` is what is sent to the device. And I also captured the console logs to see the network IP scanning Butter was doing.

Keep in mind these logs are backwards. The query order is: `Available IPs filtered by allowed network types -> Query Device IP of the appropriate network types -> Choose a local streaming URL based on what is closest to the device IP`.

So here are the results:

### `ipFamily: App.Device.IP_FAMILY.ALL` (also used when a device has NO ipFamily property at all):

```bash
generic.js:91 Device IP: 10.0.1.164
generic.js:92 Available IPs: ["fe80::14d9:e47d:1ee7:4dd9","10.0.1.248","fe80::fd76:f651:a4b9:38ed","172.16.205.1","172.16.67.1"]
generic.js:94 10.0.1.248 picked for external playback
airplay.js:45 >>>>>>THEURL:http://10.0.1.248:48047/2
airplay.js:46 AirPlay.play() streaming to "Apple TV"
```

Behavior: It loads all interfaces. Then your `options` code detects that "ALL" (both v4 and v6) are okay for this device library so it tells Node to look up "either IP type", and Node detects the device IP as IPv4, and Butter gives the device an IPv4 streaming URL.

### `ipFamily: App.Device.IP_FAMILY.IPv6`:

```bash
generic.js:91 Device IP: fe80::1021:6ea2:5654:752c
generic.js:92 Available IPs: ["fe80::14d9:e47d:1ee7:4dd9","fe80::fd76:f651:a4b9:38ed"]
generic.js:94 fe80::14d9:e47d:1ee7:4dd9 picked for external playback
airplay.js:46 >>>>>>THEURL:http://fe80::14d9:e47d:1ee7:4dd9:12931/2
airplay.js:47 AirPlay.play() streaming to "Apple TV"
```

Behavior: It filters to only IPv6 interfaces. Then your `options` code detects that the device library supports only IPv6 and does an IPv6 query (good), and Node detects the device IP as IPv6, and Butter gives the device an IPv6 streaming URL.

### `ipFamily: App.Device.IP_FAMILY.IPv4`:

```bash
generic.js:91 Device IP: 10.0.1.164
generic.js:92 Available IPs: ["10.0.1.248","172.16.205.1","172.16.67.1"]
generic.js:94 10.0.1.248 picked for external playback
airplay.js:45 >>>>>>THEURL:http://10.0.1.248:18611/2
airplay.js:46 AirPlay.play() streaming to "Apple TV"
```

Behavior: It filters to only IPv4 interfaces. Then your `options` code detects that the device library supports only IPv4 and does an IPv4 query (good), and Node detects the device IP as IPv4, and Butter gives the device an IPv4 streaming URL.

## Conclusion

Your work on the new IP selection routine is perfect! :+1 

This long-lasting IP selection bug for AppleTVs is finally solved.

I'll return discussions to the regular ticket: https://github.com/butterproject/butter-desktop/issues/528

xaiki:
cool thanks for a thourough review ! will await your patch !

PR #543, opened 2016-12-17
Fix small typo from PR #537

SteveJobzniak:
Fixes typo in #537.

https://github.com/butterproject/butter-desktop/pull/537#issuecomment-267685850

### DIFF
--- a/src/app/lib/device/generic.js
+++ b/src/app/lib/device/generic.js
@@ -37,7 +37,7 @@
             id: 'local',
             type: 'local',
             typeFamily: 'internal',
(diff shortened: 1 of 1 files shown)


xaiki:
it is not a typo it's actually a bug ;)

SteveJobzniak:
@xaiki Well, the dns lookup options id currently falling back to "all" when ipFamily isn't set to any specific fanily. So this code change does nothing to the code behavior. It just makes it look correct. ;) So I guess it's something that isn't a bug and isn't a typo. A tug? No, that sounds dirty,

PR #550, opened 2016-12-19
remove dot class make it part of the CSS

### LINKED ISSUE #516: movie details view issues
- [ ] Visually: at 960x580 (min WxH for the window), it's ugly: 
nw_2016-12-04_11-33-50

- [x] things move around when opening/closing dropdowns, they should be on top of other thing, and not make small 'jumps' in the UI

- [ ] Logs say "audio:none" but it's english. It isn't 'none'. Maybe something to set in butterproviders?

- [ ] dots should be connected to each metadata entry (year, runtime, genres, etc) so we can hide() a missing metadata element and it hides the dot at the same time

- [ ] render health is broken and is temporarily put in a try/catch

xaiki:
Signed-off-by: Niv Sardi <xaiki@evilgiggle.com>

Fixes #516 or at least parts of it.

### DIFF
--- a/src/app/styl/views/movie_detail.styl
+++ b/src/app/styl/views/movie_detail.styl
@@ -166,7 +166,8 @@
                             height 1em
                             line-height 1em
                             margin-top 2px
-                            margin-right: -3px
+                            margin-right -2px
+                            margin-left 2px
 
                 .health-icon
                     position: relative
--- a/src/app/styl/views/movie_detail.styl
+++ b/src/app/styl/views/movie_detail.styl
@@ -94,7 +94,7 @@
                     text-overflow: ellipsis
 
                     &::before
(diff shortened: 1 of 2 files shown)


vankasteelj:
There's an issue with the rating stars dot, it's underlined. I couldnt find out in devtools what was causing it

vankasteelj:
This will realign the stars, there still is the underline issue. 
```
.rating-star-half-container {
    width: 1em;
    height: 1em;
    line-height: 1em;
    margin-left: 2px;
    margin-right: -2px;
}
```

vankasteelj:
the underline is only present in vodo movie details : wtf?

xaiki:
@vankasteelj the bug was from the use of content: ".", fixed that and the alignments.

PR #552, opened 2016-12-19
switch minimize to events

### LINKED ISSUE #549: Switch "minimize in taskbar" to event instead of click
Currently, minimizing in taskbar (app.js) only works when clicking the button. It needs to be switched to on('minimize') event

vankasteelj:
- fixes #549

### DIFF
--- a/src/app/app.js
+++ b/src/app/app.js
@@ -340,6 +340,7 @@ window.ondragenter = function (e) {
 };
 
 var minimizeToTray = function () {
+
     win.hide();
     win.isTray = true;
 
@@ -678,6 +679,12 @@ nw.App.on('open', function (cmd) {
         } else if (file.endsWith('.torrent')) {
             handleTorrent(file);
         }
+    }
+});
+
+win.on('minimize', function () {
+    if (Settings.minimizeToTray) {
+        minimizeToTray();
     }
 });
 
--- a/src/app/lib/views/title_bar.js
+++ b/src/app/lib/views/title_bar.js
@@ -71,12 +71,7 @@
         },
 
         minimize: function () {
-            var that = this.nativeWindow;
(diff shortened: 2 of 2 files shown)


src/app/lib/views/title_bar.js
```diff
@@ -71,12 +71,7 @@
         },
 
         minimize: function () {
-            var that = this.nativeWindow;
-            if (AdvSettings.get('minimizeToTray')) {
-                minimizeToTray();
-            } else {
-                that.minimize();
-            }
+            this.nativeWindow.minimize();
```
xaiki: so that fires the 'minimize' event we handle later ? it looks wierd and i'm not sure i can test that on my setup (i have no tray)

vankasteelj: instead of using a minimize custom function, I intercept the on('minimize') event, that's nwjs native event. Without the modification, the only way to minimize to tray is to click on the minimize button, not any other way (native OS shortcut, clicking on icon, etc). Tray works on Ubuntu, there's no reason it wouldnt work on gnome

src/app/app.js
```diff
@@ -681,6 +682,12 @@ nw.App.on('open', function (cmd) {
     }
 });
 
+win.on('minimize', function () {
+    if (Settings.minimizeToTray) {
+        minimizeToTray();
+    }
```
xaiki: shouldn't we stop events or something here ?
just reading through that patch makes me think we're loosing functionality.... are we sure it works on windows, osx... ?

vankasteelj: if you stop the event, there will be no minimizing

xaiki:
can't really test this on all platforms, but it does look good, so merge it now.

PR #564, opened 2016-12-26
win.on{restore,maximize}

### LINKED ISSUE #560: restore maximized state on start
Currently, it somewhat tries to match the maximized size but with a few pixels off. Should be easy to fix. Bug on windows 10

vankasteelj:
- restore state on app start
- display the right icon on windows
- fixes #560

### DIFF
--- a/src/app/app.js
+++ b/src/app/app.js
@@ -645,7 +645,7 @@ win.on('move', function (x, y) {
 });
 
 win.on('enter-fullscreen', function () {
-    App.vent.trigger('window:focus');
+    win.focus();
 });
 
 win.on('minimize', function () {
--- a/src/app/app.js
+++ b/src/app/app.js
@@ -158,9 +158,9 @@ var initApp = function () {
     var mainWindow = new App.View.MainWindow();
 
     // -m argument to open minimized to tray
-    if (nw.App.fullArgv.indexOf('-m') === -1) {
-        win.show(true);
-    }
+    var isStartMinimized = nw.App.fullArgv.indexOf('-m') !== -1;
+
+    win.show(isStartMinimized);
 
     try {
         App.Window.show(mainWindow);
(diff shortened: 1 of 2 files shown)


src/app/app.js
```diff
@@ -656,6 +633,58 @@ if (nw.App.fullArgv.indexOf('-f') !== -1) {
...
+win.on('minimize', function () {
+    if (Settings.minimizeToTray) {
+        minimizeToTray();
+    }
+});
+
+win.on('maximize', function () {
+    localStorage.isMaximized = 1;
+    if (process.platform === 'win32') {
+        $('.os-max').addClass('os-is-max');
```
xaiki: why only on windows ?

src/app/app.js
```diff
@@ -656,6 +633,58 @@ if (nw.App.fullArgv.indexOf('-f') !== -1) {
...
+    }
+});
+
+win.on('restore', function () {
+    if (Boolean(parseInt(localStorage.isMaximized))) {
```
xaiki: is it expensive to set it anyhow ?

src/app/app.js
```diff
@@ -153,7 +159,7 @@ var initApp = function () {
 
     // -m argument to open minimized to tray
     if (nw.App.fullArgv.indexOf('-m') === -1) {
-        win.show();
+        win.show(true);
```
xaiki: what does that fix ?

src/app/app.js
vankasteelj: restore is fired both on unminimize (deprecated) and unmaximize (deprecated)

src/app/app.js
vankasteelj: because other use the dot thingy, at least that's what I remember. Maybe os-is-max won't hurt them, and we can set it anyway. Idk

src/app/app.js
vankasteelj: remove the lines below "on('focus', nanah)"

src/app/app.js
xaiki: i think it's not the right place for that check, set os-is-max and we can fix it in the CSS later.

sopex: yes I agree with @xaiki

PR #567, opened 2016-12-27
Btzr/redesign

abadfederico:


xaiki:
please PR against `xaiki/redesign`

PR #574, opened 2016-12-29
re-add ghosts, done right

### LINKED ISSUE #562: item rows not justified
A picture is worth a thousand words:

croquis

vankasteelj:
- fixes #562

Proposal. We add 10 items no matter what: their height is 0 so we don't care.

### DIFF
--- a/src/app/lib/views/browser/list.js
+++ b/src/app/lib/views/browser/list.js
@@ -262,9 +262,9 @@
 
         onLoaded: function () {
             App.vent.trigger('list:loaded');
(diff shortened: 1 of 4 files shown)


vankasteelj:
this is actually not cached. wtf.

vankasteelj:
from my tests, we have gained between 2 and 12ms (on my machine) per 'onLoaded' call, even though we now add ghosts on top of loadMore.

xaiki:
please rebase this into `xaiki/redesign` or `btzr/redesign`

PR #583, opened 2017-01-02
Vkj/kb nav

vankasteelj:
it's... weird

### DIFF
--- a/src/app/lib/views/browser/list.js
+++ b/src/app/lib/views/browser/list.js
@@ -90,7 +90,7 @@
             Mousetrap.bind(['up', 'down', 'left', 'right'], this.move.bind(this));
             Mousetrap.bind('f', this.toggleSelectedFavourite.bind(this));
             Mousetrap.bind('w', this.toggleSelectedWatched.bind(this));
-            Mousetrap.bind(['enter', 'space'], this.selectItem.bind(this));
+            Mousetrap.bind(['enter', 'space'], this.clickItem.bind(this));
             Mousetrap.bind(['ctrl+f', 'command+f'], this.focusSearch.bind(this));//FIXME: needs to be moved elsewhere
             Mousetrap(document.querySelector('input')).bind(['ctrl+f', 'command+f', 'esc'], this.blurSearch.bind(this));//FIXME: needs to be moved elsewhere
             Mousetrap.bind(['tab', 'shift+tab'], this.switchTab.bind(this));//FIXME: needs to be moved elsewhere
@@ -320,23 +320,14 @@
             }
         },
 
-        selectItem: function (e) {
+        clickItem: function (e) {
             e.stopPropagation();
 
             $('.item.selected .cover').click();
         },
 
-        selectIndex: function (index) {
-            var items = $('.items .item');
-
-            if (index >= items.length) {
-                index = items.length - 1;
-            }
-
-            var next = items.eq(index);
-            var previous = $('.items .item.selected');
-
-            previous.removeClass('selected');
+        selectItem: function (prev, next) {
+            prev.removeClass('selected');
             next.addClass('selected');
 
             if (!Common.isElementInViewport(next)) {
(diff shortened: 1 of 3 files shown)


src/app/lib/views/browser/list.js
```diff
@@ -386,7 +394,7 @@
             if (index === -1) {
                 index = 0;
             } else {
-                index = index + NUM_ITEMS_IN_ROW;
+                index = index + this.itemsPerRow();
```
xaiki: can we merge the 2 last functions ? so we don't have, for instance, limit case working on Up but not on Down ?

src/app/lib/views/browser/list.js
```diff
@@ -273,7 +280,7 @@
                     '</div>' +
                 '</div>';
 
-            var ghosts = '<div class="ghost"></div>'.repeat(10);
+            var ghosts = '<div class="ghost"></div>'.repeat(this.itemsPerRow()*2);
```
xaiki: so i guess the *2 is to support not having glitches when resizeing from a much lower resolution, but that assumes that our max res is not bigger than 2x our smallest res...

i'm not sure it's a safe bet to take ? the correct value should be screensize(), but then what happens if we move the app to another screen that is bigger ?

src/app/lib/views/browser/list.js
```diff
@@ -237,25 +233,36 @@
...
-            var total = 0;
-            var items = $(document.querySelector('.items')).children('.item');
+            var currentWidth = this.$el.width();
+            var minPerRow = Math.floor(1 / (Settings.postersWidth / currentWidth) - 1);
+
+            // recalc only if needed
+            if (
+                currentWidth !== this.ui.row.listWidth // win was resized?
+                || Settings.postersWidth !== this.ui.row.posterWidth // poster were resized?
+                || this.ui.row.itemsPerRow < minPerRow // something impossible happened?
```
xaiki: returning early here makes the code more readable and introduces less whitespace changes.

src/app/lib/views/browser/list.js
```diff
@@ -237,25 +233,36 @@
         },
 
         itemsPerRow: function () {
-            var total = 0;
-            var items = $(document.querySelector('.items')).children('.item');
+            var currentWidth = this.$el.width();
+            var minPerRow = Math.floor(1 / (Settings.postersWidth / currentWidth) - 1);
```
xaiki: i discovered that you can use ~~ for floor =) i can't stop using it since then.

xaiki: it's `~~`

src/app/lib/views/browser/list.js
vankasteelj: we're limited in our postersize, so *2 is a safe bet. I'll check if I can do something with minPerRow, it might be better

src/app/lib/views/browser/list.js
vankasteelj: I'll have to check

vankasteelj:
Note that it also improves the experience, because now we can even hit 'down' wherever we want, and it will go down to the last item if no item directly under exists

xaiki:
ping ?

src/app/lib/views/browser/list.js
vankasteelj: I disagree. It would duplicate the return and would change the conditions to && making it actually harder to process, because all conditions will need to be checked.

Test conditions yourself: 
```
var a = function (s) {console.log('fn a'); return s;};
var b = function (s) {console.log('fn b'); return s;};
var c = function (s) {console.log('fn c'); return s;};
if (!a(false) || !b(false) || !c(false)) console.log('done');
```

xaiki: then wrap the calculations in a function

src/app/lib/views/browser/list.js
```diff
@@ -370,61 +346,22 @@
...
-            this.selectIndex(index);
-        },
-
-        moveLeft: function (e) {
+        move: function (e) {
             e.preventDefault();
             e.stopPropagation();
 
+            var row = this.itemsPerRow();
             var index = $('.items .item.selected').index();
```
xaiki: you promised no more jQuery !

src/app/lib/views/browser/list.js
```diff
@@ -370,61 +346,22 @@
...
-            if (index === -1) {
-                index = 0;
-            } else if (index === 0) {
-                index = 0;
-            } else {
-                index = index - 1;
-            }
-            this.selectIndex(index);
-        },
+            index < 0 && (index = 0);
```
xaiki: isn't it missing a check on the other bound ? i.e. index > numItems ?

xaiki: i'd rather have a check for `index === -1` as what we're checking is if we had a match

src/app/lib/views/browser/list.js
```diff
@@ -370,61 +346,22 @@
...
+            };
 
-            var index = $('.items .item.selected').index();
-            if (index === -1) {
-                index = 0;
-            } else {
-                index = index + 1;
-            }
-            this.selectIndex(index);
+            this.selectIndex(Math.max(0, map[e.key]));
```
xaiki: we're missing a check on the other bound, i.e. min(numitems, newindex)

vankasteelj: `Math.max(0, n)` is to get 0 if `n` is negative.

xaiki: yes, what about if n is > numberElements ?

vankasteelj: that's handled in selectIndex (was, actually, see new commit)

PR #586, opened 2017-01-03
remote debug on control+shift+i

vankasteelj:
- port is set in package.json


xaiki:
please PR into `xaiki/redesign` that's where most dev is done today.

xaiki:
cherry picked into `xaiki/redesign`

PR #589, opened 2017-01-07
allow resizing player below the 1/4 FHD limit

### LINKED ISSUE #588: Allow to resize player to lower size
While I'm all in for the app size being locked to a minimum to spare us the difficulty of coding fully responsive css, I found myself in need of a smaller player than 960x520.

What about, on player open:
  
     win.setMinimumSize(500,300)

And reset on player close.

.player-title width also has to be adjusted. calc(100% - 160px) seems to do the trick.

vankasteelj:
- fix #588
- restores the original state after player close

### DIFF
--- a/src/app/lib/views/player/player.js
+++ b/src/app/lib/views/player/player.js
@@ -353,12 +353,32 @@
                 }
             }
         },
+        
+        allowResize: function (state) {
+            if (state) {
+                this.appPreviousState = {
+                    width: win.width,
+                    height: win.height,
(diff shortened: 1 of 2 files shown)


xaiki:
I don't really like this, it makes jumping size, what do you say @abadfederico ?

vankasteelj:
Ah yes, I should prevent resizing if the player didn't go below 'default' values. The goal is only to resize if the player was set under 960x520.

vankasteelj:
updated

PR #593, opened 2017-01-09
Xaiki/abad

xaiki:
resync the show-view redesign work.
still WIP.


vankasteelj:
noooooooooooooo

src/app/app.js
```diff
@@ -162,11 +162,7 @@ var initApp = function () {
 
     win.show(isStartMinimized);
 
-    try {
-        App.Window.show(mainWindow);
-    } catch (e) {
-        console.error('Couldn\'t start app: ', e, e.stack);
-    }
+    App.Window.show(mainWindow);
```
vankasteelj: You sure about this?

src/app/index.html
```diff
@@ -46,11 +48,11 @@
 
 	<!-- 3rd party Dependencies -->
 	<script src="vendor/jquery/dist/jquery.min.js"></script>
-	<script src="vendor/underscore/underscore-min.js"></script>
-	<script src="vendor/backbone/backbone-min.js"></script>
-	<script src="vendor/backbone.wreqr/lib/backbone.wreqr.min.js"></script>
-	<script src="vendor/backbone.babysitter/lib/backbone.babysitter.min.js"></script>
-	<script src="vendor/marionette/lib/backbone.marionette.min.js"></script>
+	<script src="vendor/underscore/underscore.js"></script>
```
vankasteelj: why not minified?

src/app/lib/models/content_item.js
```diff
@@ -12,7 +12,7 @@
         initialize: function (attrs) {
             var providers = Object.assign(this.getProviders(), attrs.providers);
             this.set('providers', providers);
-            this.set('idAttribute', this.idAttribute);
+            this.set('idAttribute', attrs.uniqueId || this.get('uniqueId') || this.idAttribute);
```
vankasteelj: dafuq is this? I thought there was a defined api?

src/app/lib/views/generic_detail.js
```diff
@@ -0,0 +1,238 @@
+(function(App) {
+    'use strict';
+
+    var loadImage = function (img, type, onLoad) {
```
vankasteelj: I don't like having a floating variable on top of the file

src/app/lib/views/generic_detail.js
```diff
@@ -0,0 +1,238 @@
...
+            App.MovieDetailView = this;
+
+            this.loadImages();
+            this.loadComponents();
+
+            $('.spinner').hide();
+        },
+
+        loadComponents: function () {
+            console.error('implement in subclasses');
```
vankasteelj: dafuq?

src/app/lib/views/main_window.js
```diff
@@ -99,11 +99,11 @@
 
             // Movies
             App.vent.on('movie:showDetail', _.bind(this.showMovieDetail, this));
-            App.vent.on('movie:closeDetail', _.bind(this.closeMovieDetail, this.MovieDetail));
+            App.vent.on('movie:closeDetail', _.bind(this.closeMovieDetail, this.DetailsContainer));
 
             // Tv Shows
             App.vent.on('tvshow:showDetail', _.bind(this.showShowDetail, this));
-            App.vent.on('tvshow:closeDetail', _.bind(this.closeShowDetail, this.MovieDetail));
+            App.vent.on('tvshow:closeDetail', _.bind(this.closeShowDetail, this.DetailsContainer));
```
vankasteelj: why not a closeDetail event, without specifying the type?

src/app/lib/views/movie_detail.js
```diff
@@ -1,220 +1,19 @@
...
-            'click .movie-imdb-link': 'openIMDb',
-            'mousedown .magnet-link': 'openMagnet',
-            'click .rating-container': 'switchRating'
-        },
-
+    App.View.MovieDetail = App.View.GenericDetail.extend({
         regions: {
-            PlayControl: '#play-control'
+            ActionBar:  '#action-bar',
+            DetailMeat: '#detail-meat'
```
vankasteelj: I have no idea why meat. It's an ugly word, like in "wanna see my meat?"

src/app/styl/Official_-_Experimental_theme.styl
```diff
@@ -39,8 +39,8 @@ $Pure_White = #FFFFFF
     $FullScreenButtonHover = $Pure_White
 
     $FilterBarBg = $NavColor
-    $FilterBarText = #c3c5c7
```
vankasteelj: then there is no point in having a hover color if you set the same for both. UX is better, imo, with changes when you hover. Especially in this case where filterbar is not important and the attention should be dragged below, to the list

src/app/styl/views/filter_bar.styl
```diff
@@ -24,16 +24,18 @@
         font-size: 1.4em
         text-transform: uppercase
         border-bottom: 3px solid transparent
+        opacity: 0.4
 
         &:hover
-            color: $FilterBarTextHover
+            opacity: 0.9
```
vankasteelj: Ok oh, I see. But then you're forcing a set opacity, instead of leaving a theming choice

src/app/styl/views/generic_detail.styl
```diff
@@ -0,0 +1,36 @@
...
+        position: absolute
+        opacity: 0
+        background-position-x 50%
+        background-position-y -33px
+        -webkit-filter: blur(8px) brightness(0.7)
+        background-repeat: no-repeat
+        background-size: cover
+
+        &.fadein
+            transition: opacity .3s ease-in;
```
vankasteelj: transition should be set on backdrop, not fadein

src/app/templates/movie-detail.tpl
```diff
@@ -1,77 +1,50 @@
 <%
-if(typeof health === "undefined"){ health = false; };
 if(typeof synopsis === "undefined"){ synopsis = "Synopsis not available."; }; 
 if(typeof runtime === "undefined"){ runtime = "N/A"; };
-if (genre) {
-    for(var i = 0; i < genre.length; i++) {
-        genre[i] = i18n.__(genre[i]); 
-    }
+if (genres) {
```
vankasteelj: typeof should be used instead of if(), because if() will throw an error where typeof wont

src/app/templates/show-header.tpl
```diff
@@ -0,0 +1,31 @@
...
+        </a>
+        <div id="collapseOne" class="panel-collapse collapse in" role="tabpanel" aria-labelledby="headingOne">
+        </div>
+    </li>
+    <%_.each(torrents, function(episodes, season) { %>
+    <li class="panel">
+        <a class="source" role="button" data-toggle="collapse" data-parent="#accordion"
+           href="#show-detail-collapse-<%= season %>"
+           aria-expanded="false" aria-controls="show-detail-collapse-<%= season %>">
+            <%= i18n.__('Season ' + season) %>
```
vankasteelj: this looks wrong. `i18n.__('Season') + season` should be set

src/app/templates/show-header.tpl
```diff
@@ -0,0 +1,31 @@
...
+        </a>
+        <ul id="show-detail-collapse-<%= season %>" class="episode-list panel-collapse collapse" role="tabpanel">
+            <%_.each(episodes, function(value, episode) { %>
+            <li>
+                <span><%= episode %></span>
```
vankasteelj: no i18n?

src/app/lib/views/movie_detail.js
xaiki: yeah, suggestions ? i ran out of name, it's really detail-detail

src/app/app.js
xaiki: i couldn't find a situation where it helpped

src/app/index.html
xaiki: i was debugging things in underscore and backbone, minified made my life hell.

src/app/lib/models/content_item.js
xaiki: so we were getting into that code by many different paths... and Backbone init does weird things on some of those paths...

src/app/lib/views/movie_detail.js
vankasteelj: #details?

src/app/templates/show-header.tpl
xaiki: what about languages where you say '0 seasonea' ?

src/app/templates/show-header.tpl
xaiki: it's a number.

src/app/lib/views/main_window.js
xaiki: that couuuuld work...

src/app/lib/views/generic_detail.js
xaiki: we implement that in MovieDetail and ShowDetail

src/app/templates/show-header.tpl
vankasteelj: then i18n.__('Season %s', season), but no +. You're screwing the entire point of translation with a +, because the string is modified before being sent to i18n, which will then try to translate 'Season 18' instead of 'Season' and a number.

src/app/lib/views/movie_detail.js
xaiki: yeah, but then it's called 'generic-detail', and we have 'detail', it's weird isn't it ?

src/app/lib/views/generic_detail.js
vankasteelj: Yes, but if it is implemented, there is is no error. The error should be triggered by the caller, not the callee

src/app/app.js
vankasteelj: me neither, but still :D

src/app/lib/views/main_window.js
xaiki: it's actually requiring a bit more work...

src/app/templates/movie-detail.tpl
xaiki: how can if throw an error ? it's a variable ?

src/app/lib/views/generic_detail.js
xaiki: well it *needs* *to* be implemented, so this is just overwritten.

xaiki: and there is no caller/callee this is subclassing.

src/app/templates/movie-detail.tpl
vankasteelj: if `genres` is undefined, trying to see if it exists will throw an error. Try `if (heuzqejhq) {}`

src/app/styl/views/generic_detail.styl
btzr-io: fixed

src/app/lib/views/generic_detail.js
vankasteelj: No, what I'm saying is that this function does absolutely nothing apart displaying an error (not even triggering one) that would not be logged if the function wasnt here in the first place. I don't see a scenario where this function is useful here. You say yourself it's implemented in subclasses.

xaiki: it's there so that you notice if you forget to implement it in subclasses.

vankasteelj:
- movie details & show details open to a blank view
- settings are fucked up

PR #605, opened 2017-02-13
Increase maxBuffer size for parseReqDeps

chtrinh:
'npm ls' stdout can execeed max buffer size of 200KB.
Setting child process max buffer size to 500KB.


xaiki:
this will actually fix a heap of build issues.

PR #606, opened 2017-02-13
Remove mouseScroll listner event

chtrinh:
`.bind()` returns a new function which is not referenced
when trying to removed later. Saving bounded function as a property
to be referenced.

### DIFF
--- a/src/app/lib/views/player/player.js
+++ b/src/app/lib/views/player/player.js
@@ -36,12 +36,14 @@
             this.listenTo(this.model, 'change:active_peers', this.updateActivePeers);
             this.listenTo(this.model, 'change:downloaded', this.updateDownloaded);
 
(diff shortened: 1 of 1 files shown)


xaiki:
looks good to me, but i'll let @vankasteelj review this one, he knows better the code.

vankasteelj:
it seems ok. I don't know enough about events to have something to say about this, it does work, and it removes a jquery call. I'm up for that. And if it indeed correctly unbinds the event, it's even better.

chtrinh:
If you start the player and then close it. Scrolling right afterwards
trigger an error message in your dev tools

vankasteelj:
that's what that message was? lol

vankasteelj:
I thought it was javascript fucking up, simply :D

chtrinh:
@vankasteelj: 😆 , i thought so too!

PR #617, opened 2017-03-03
Decrease min width to 480

matthewbauer:
Things seem to work fine with this. It could probably be even smaller though.

- min width set to 480


vankasteelj:
It doesnt work on details for shows/movie

btzr-io:
That should be fixed in the new layout ^^

xaiki:
This patch is simple enoegh let,s push it no the redesign

PR #649, opened 2017-06-03
removed unworking line which I believe tried to focus on the first re…

Sioma112233:
## Fixes
fixed issue where after searching for a movie/show/etc hitting CTRL+F would result on focusing on the search-bar for only a split second (then the focus would move the `$(".items")` div which would not do anything (if focusing on the first result after search was meant in that line of code, it did not happen, it would focus on the parent div which does nothing but interfere with the search bar focus).

removing that focus line solves the focusing on the search bar.

### notes for reproducing the bug
- do not use the mouse at any point (focus should stay on that `$(".items")` div)
- hit CTRL+F and search something
- after u got the results (or no results at all), try hitting CTRL+F again and searching for something else

### DIFF
--- a/src/app/lib/views/browser/list.js
+++ b/src/app/lib/views/browser/list.js
@@ -286,7 +286,6 @@
             $('.items').attr('tabindex', '1');
             _.defer(function () {
                 self.checkFetchMore();
-                self.$('.items:first').focus();
             });
 
         },


xaiki:
we have a problem, but i had a run at your branch and it doesn't seem to fix it.

xaiki:
that said, the toolbar will probably get soonishly rewritten in the `xaiki/react` branch (as a react component)

Sioma112233:
Are u sure, I also asked a friend to check it, and it said it worked for him.
Not familiar with the project, but maybe u have some cached version that's running instead or something like that?
Even if you're rewritting that view in the near month, it's better to have a temporary fix now no?

xaiki:
@vankasteelj want to have a look at this ?

vankasteelj:
it indeed does not work, but what's the fix?

vankasteelj:
```
        selectFirstItem: function () {
            var firstItem = $('.items').find('.item:visible:first');
            this.selectItem(null, $(firstItem));
        },

        selectItem: function (prev, next) {
            prev && prev.removeClass('selected');
            next.addClass('selected');

            if (!Common.isElementInViewport(next)) {
                next[0].scrollIntoView(false);
                this.onScroll();
            }
        },
```

and `this.selectFirstItem();` added to onLoaded

vankasteelj:
could just be : 

```
        selectFirstItem: function () {
            this.selectItem(null, $('.item:first'));
        },
```

but then you can't hide anything, you have to DELETE the items. And who knows what tomorrow holds, we might want to display:none some items

vankasteelj:
Also, don't pay attention to what I wrote if you're on master, it's outdated (severely regarding list.js)

xaiki:
this has been fixed in our dev branch `xaiki/react` will be fixed on master when we merge

PR #654, opened 2017-06-19
fix butter-component-settings npm 'invalid' error

vankasteelj:


VitorVRS:
butter-component-settings is published at npm with dist folder.
Using dependency directly from github are not working, unless i run ```npm run build``` inside butter-component-settings

xaiki:
building should be part of the install, that's a `-settings` issue, can you open it there @VitorVRS ?

PR #672, opened 2017-10-24
add more hebrew tranlation

asafofer:
Fixes # .

- 
- 
- 


xaiki:
@insertc0in

xaiki:
I don't want to lose this, but i think it should go to transifex.

insertc0in:
@xaiki I'll sync this up with Transifex. In the meantime, the source language (English) has an integer value in https://github.com/butterproject/butter-desktop/blob/master/src/app/language/en.json#L515 - Transifex only accepts strings. I don't know if that was intentional or not, but it should be string and not integer.

insertc0in (approved):

PR #694, opened 2018-03-28
Newer dependencies. It doesn't break it :)

sopex:
Fixes # .

- 
- 
- 


xaiki:
it looks like this patch is for `xaiki/react` (that we should merge into master)

